PR #52, opened 2023-06-14
style: modify pkg about database connection

whhe:
Move common process of database connection to 'pkg/util/database', three main changes:
- `pkg/oceanbase/connector/ob_connect_properties.go` -> `pkg/oceanbase/connector/datasource.go`
- `pkg/oceanbase/connector/ob_connector.go` -> `pkg/util/database/connector.go`
- `pkg/oceanbase/connector/ob_connector_manager.go` -> `pkg/util/database/pool.go`


pkg/util/database/connector.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	// register mysql driver
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/jmoiron/sqlx"
+)
+
+// Connector represents a connection pool.
+type Connector struct {
+	ds     DataSource
+	client *sqlx.DB
```
chris-sun-star, line 24: The original implementation supports connection pool config of client, which means a connector is able to handle multiple db connections, this is useful when multiple goroutines use the same connector

pkg/util/database/connector.go
```diff
@@ -0,0 +1,70 @@
+/*
```
chris-sun-star, line 1: database can be a package under pkg, no need to put it under util

whhe: Done.

chris-sun-star (approved):
LGTM

PR #66, opened 2023-09-19
optimize cluster operation and some other fixes

chris-sun-star:
## Summary

1. add task recovery
2. add more integration test for obcluster operation
3. automatically alter sys tenant's locality and zonelist when zone operation
4. check observer indeed exists in obcluster after recover, if not add an add observer process
5. maintain parameters only when obcluster is in running status
6. create service for monitor service discovery
7. fix data race when update status and add a parameter when build docker to enable race detection
8. grant more resource to manager container


deploy/obcluster.yaml
```diff
@@ -14,16 +14,10 @@ spec:
   topology:
     - zone: zone1
       replica: 1
-      nodeSelector:
```
PowerfooI, line 17: Don't we need nodeSelector any more?

pkg/oceanbase/const/sql/cluster.go
```diff
@@ -13,6 +13,6 @@ See the Mulan PSL v2 for more details.
 package sql
 
 const (
-	BootstrapServer = "REGION '%s' ZONE '%s' SERVER '%s:%d'"
```
PowerfooI, line 16: Why delete REGION here?

pkg/oceanbase/operation/server.go
```diff
@@ -21,12 +21,17 @@ import (
...
-	err := m.QueryRow(observer, sql.GetServer, s.Ip, s.Port)
+	observers, err := m.ListServers()
 	if err != nil {
-		return nil, errors.Wrap(err, "Get observer failed")
+		return nil, err
 	}
-	return observer, nil
+	for _, observer := range observers {
+		if s.Ip == observer.Ip && s.Port == observer.Port {
+			return &observer, nil
```
PowerfooI, line 30: Why do we list all servers here instead query exactly?

pkg/resource/obcluster_task.go
```diff
@@ -141,11 +143,12 @@ func (m *OBClusterManager) ModifyOBZoneReplica() error {
...
 }
 
-func (m *OBClusterManager) DeleteOBZone() error {
+func (m *OBClusterManager) getZonesToDelete() ([]v1alpha1.OBZone, error) {
+	deletedZones := make([]v1alpha1.OBZone, 0)
 	obzoneList, err := m.listOBZones()
 	if err != nil {
 		m.Logger.Error(err, "List obzone failed")
-		return errors.Wrapf(err, "List obzone of obcluster %s failed", m.OBCluster.Name)
+		return deletedZones, errors.Wrapf(err, "List obzone of obcluster %s failed", m.OBCluster.Name)
```
PowerfooI, line 151: why not return (nil, err) here :)

pkg/resource/observer_task.go
```diff
@@ -634,7 +635,7 @@ func (m *OBServerManager) WaitOBServerDeletedInCluster() error {
 			return errors.Wrapf(err, "Get oceanbase operation manager failed")
 		}
 		observer, err := operationManager.GetServer(observerInfo)
-		if observer == nil {
+		if observer == nil && err == nil {
```
PowerfooI, line 638: What if `err != nil`, shall we log the error information?

deploy/obcluster.yaml
chris-sun-star: node selector is optional，you can use it to specific certain nodes to run the observer pods, if it's not set, it depends on the K8s scheduler, since this is an example, I'd like to make as simple as possible.

pkg/oceanbase/const/sql/cluster.go
chris-sun-star: region is not required, and it can be set after the cluster starts

pkg/oceanbase/operation/server.go
chris-sun-star: If query the exactly record, it will return an error when the record is not exits, so I make it return nil, nil instead of return an error for record not found scenario.

pkg/resource/obcluster_task.go
chris-sun-star: this function returns []v1alpha1.OBZone, err, not *[]v1alpha1.OBZone, err

pkg/resource/observer_task.go
chris-sun-star: ok, I'll add log for error

pkg/oceanbase/operation/server.go
PowerfooI: It would be better that we add `where ip = ? and port = ?` filter when listing observer, we can get empty slice if no servers matching the condition without return an error. Listing all servers and loop them may become an overhead perhaps.

pkg/resource/obcluster_task.go
PowerfooI: yes, slice type is actually a pointer type, it should be ok to return `nil`. The following test case passes.

```go
It("Empty Slice", Label("misc"), func() {
	var a []int
	Expect(len(a)).To(BeEquivalentTo(0))
	Expect(a).To(BeEmpty())
	Expect(a).To(BeNil())
})
```

PowerfooI:
LGTM

PR #69, opened 2023-09-20
optimize: controller requeue config, proper status update, data query

chris-sun-star:
## Summary

1. update status with retry
2. remove global sync config, config event requeue policy by each controller, and gives default by coordinator
3. add timeout method to query row and query list and gives default value


PowerfooI (commented):
LGTM 💯

PR #72, opened 2023-09-22
fix tenant add replica, optimize resource update with retry

chris-sun-star:
## Summary

1. fix tenant add replica issue
2. optimize resource update with retry
3. support specify for build observer image script


PowerfooI (commented):
LGTM

PR #87, opened 2023-10-25
Feature: configure togological structure with affinity and taint tolerations

PowerfooI:
## Summary

Feature: configure togological structure with affinity and taint tolerations
* Add fields `Affinity` and `Tolerations` that are from k8s corev1 api group to `OBZoneTopology` type
* When an OBZone creates an OBServer, it pass these topology configurations down to observer pod spec
* Along with previous `nodeSelector` field, now we have all three methods to control observer pod scheduling behaviors


chris-sun-star (approved):
LGTM

PR #88, opened 2023-10-25
Feature: helm chart to deploy cluster quickly

PowerfooI:
## Summary

Helm chart to deploy oceanbase cluster which auto generate required user secrets.

### DIFF
--- a/charts/oceanbase-cluster/Chart.yaml
+++ b/charts/oceanbase-cluster/Chart.yaml
@@ -1,0 +1,24 @@
+apiVersion: v2
+name: oceanbase-cluster
+description: A Helm chart for Kubernetes to deploy OceanBase cluster
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 4.2.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "4.2.0.0-101000032023091319"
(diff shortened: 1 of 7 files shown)


charts/oceanbase-cluster/Chart.yaml
```diff
@@ -0,0 +1,24 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 4.2.0
```
chris-sun-star, line 18: maybe this version is too short, keep this version the same as observer version

charts/oceanbase-cluster/templates/obcluster.yaml
```diff
@@ -0,0 +1,44 @@
...
+    resource:
+      {{- toYaml .resource | nindent 6 }}
+    storage:
+     {{- range $key, $size := .storages }} 
+      {{ $key }}:
+        storageClass: {{ $.Values.storageClass }}
+        size: {{ $size }}
+     {{- end }}
+    {{- end }}
+  monitor:
```
chris-sun-star, line 29: monitor is also optional, add a parameter to specific whether to enable monitor but default true

charts/oceanbase-cluster/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
```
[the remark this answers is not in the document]
PowerfooI: This api version is used by Helm. In Helm3, the field should be `v2`, it specify the helm version.

charts/oceanbase-cluster/Chart.yaml
PowerfooI: This field should follow **semver** standard, but our 4.2.0.0-YYYYMMDDHHMMss format is invalid in such style.

charts/oceanbase-cluster/Chart.yaml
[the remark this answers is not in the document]
PowerfooI: image

charts/oceanbase-cluster/templates/obcluster.yaml
PowerfooI: OK, I'll do that ~

charts/oceanbase-cluster/Chart.yaml
PowerfooI: 4.2.0-YYYYMMDDHHMMss is correct, but I think the forth number can not be ignored.

PowerfooI: or `4.2.0-subpatch.0-101000032023091319`. Major version pattern in semver is "X.Y.Z" which contains three parts, the forth number should be transformed to another form.

charts/oceanbase-cluster/Chart.yaml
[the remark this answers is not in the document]
chris-sun-star: ok, then oboperator chart also needs update api to v2

chris-sun-star (approved):

PR #89, opened 2023-10-26
Fix some tiny bugs

PowerfooI:
## Summary

1. Restoring tenant with specific `until` field will failed due to wrong field name
2. Cut oceanbase-cluster chart's default resource usage properly
3. Distribution building helpers


chris-sun-star (approved):
LGTM

PR #92, opened 2023-11-06
Fix: old connection of backup reconciler

PowerfooI:
## Summary

Reuse database connection in OBTenantBackup Reconciler by mistake, which will cause connection error if the OBCluster redeployed

## Solution Description

Remove the connection cache in OBTenantBackup Reconciler.


PowerfooI (commented):
Reviewed

PR #95, opened 2023-11-13
Bugfix: concurrent map writes

PowerfooI:
## Summary

* OBServer controller has 9 concurrent reconciles at most, so it's possible that two go routines reconcile observer at the same time. When the two (or more) go routines write the same key in task map in TaskManager which has no guarding Lock, program crashes.

## Solution Description

1. Change maps used in TaskManager from standard map to sync.Map, which is thread-safe.
2. Fix GlobalWhiteList used in OBTenantManager by the way.

### DIFF
--- a/pkg/resource/globalWhiteList.go
+++ b/pkg/resource/globalWhiteList.go
@@ -1,0 +1,17 @@
+/*
+Copyright (c) 2023 OceanBase
+ob-operator is licensed under Mulan PSL v2.
+You can use this software according to the terms and conditions of the Mulan PSL v2.
+You may obtain a copy of Mulan PSL v2 at:
+         http://license.coscl.org.cn/MulanPSL2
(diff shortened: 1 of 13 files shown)


pkg/resource/globalWhiteList.go
```diff
@@ -0,0 +1,17 @@
...
+EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+See the Mulan PSL v2 for more details.
+*/
+
+package resource
+
+import "sync"
+
+var GlobalWhiteListMap sync.Map
```
chris-sun-star, line 17: a global variable under resource package, it's better to call it GlobalTenantWhiteListMap, it only used by tenant manager

PowerfooI: OK, I renamed it to tenantWhiteListMap

chris-sun-star (approved):
LGTM

PR #98, opened 2023-11-14
Fix: annotations mask

PowerfooI:
## Summary

* Set annotations of object to nil, avoiding possible sensitive telemetry frame
* Make OBZone and OBCluster controller listen events of their owned resources. OBCluster controller listens events of OBZones, and OBCluster controller reacts when obzones change.


chris-sun-star (approved):
LGTM

PR #99, opened 2023-11-15
Optimize coordinator and fix logger setting

PowerfooI:
## Summary

1. Coordinator speeds up reconciling if status of a resource changes. Implementing a GetStatus() method for all ResourceManagers
2. Fix not working global logger setting


chris-sun-star (approved):
LGTM

PR #125, opened 2023-11-28
fix: optimize backup jobs path and omit annotation and label record

powerfooI:
## Summary

1. fix parameters in debug.mk
2. set resource labels and annotations nil
3. skip recording backup jobs


chris-sun-star (approved):
LGTM

Issue #128, opened 2023-11-30
[Enhancement]: Publish some useful packages that others could take to develop OB tools


PowerfooI:
### Description

There are some relatively independent packages like oceanbase sdk, task manager and resource coordinator that look useful to develop other tools of OceanBase. But now, they are highly coupled with other internal functionalities. It would be better if they can be purified as individual packages.

chris-sun-star:
we have made our first step
next we will consider the api definitions

PR #129, opened 2023-11-30
Refactor packages and separate `internal` and `pkg`

PowerfooI:
## Summary

1. Move auxiliary type definitions from `api/v1alpha1` to `api/types` like `BackupVolumeSpec`.
2. Move task-related types into `pkg/task` like `OperationContext`.
3. Move `controller`, `resource`, `telemetry` and `const` folder in previous `pkg/` to a new folder `internal/`.
4. In `resource` folder, the PR separate every resource manager into single folder, in which names of their tasks and task flows are included. This change increases locality of code of every resource manager and make `pkg/task` a quite independent package that does not rely on `ob-operator` actually. Every resource manager register their task flows in their own package `init` part.
5. Move struct `coordinator` and interface `ResourceManager` to a new package `coordinator` in `pkg/` folder, which is another independent module, too.
6. In order to avoid possible race when registering task flows, a simple mutex was added to `task_manager`. I believe that it does not impact the functionality of task registry. This is the only change that impacts functionalities in this PR.
7. Set specific types to fields (TaskName, FlowName, TaskID etc.) in `pkg/task` instead of plain string that may cause confusion in some cases. (e.g. mismatch name of task flows and tasks).

This PR may somehow reach the requirement of  #128


example/webapp/obproxy.yaml
```diff
@@ -45,11 +45,11 @@ spec:
             - name: OB_CLUSTER
               value: metadb
             - name: CONFIG_URL
-              value: 'http://svc-ob-configserver.oceanbase.svc:8080/services?Action=GetObProxyConfig'
+              value: 'http://svc-ob-configserver.default.svc:8080/services?Action=GetObProxyConfig'
```
chris-sun-star, line 48: wrong namespace

chris-sun-star (approved):
LGTM

PR #146, opened 2023-12-26
Rescue trapped resources controlled by ob-operator

### LINKED ISSUE #139: [Enhancement]: A method to correct status of CRs when the CRs are trapped in strange corners
### Description

Every CR has both `status.status` and `status.operationContext` fields and they may become uncontrolled after some specific operations of users. It would be better if ob-operator offers a method to correct the fields like a `op` CR which contains Group, Kind, Name, status.status and status.operationContext and it update the status of target resource.

PowerfooI:
## Summary

Add a new api type `OBResourceRescue` which can delete finalizers or reset status of target resource. So I called it `Rescue`.

fix #139


api/v1alpha1/obresourcerescue_types.go
```diff
@@ -0,0 +1,74 @@
...
+	TargetGV      string `json:"targetGV,omitempty"`
+	Namespace     string `json:"namespace,omitempty"`
+	TargetStatus  string `json:"targetStatus,omitempty"`
+}
+
+// OBResourceRescueStatus defines the observed state of OBResourceRescue
+type OBResourceRescueStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	OperationContext *tasktypes.OperationContext `json:"operationContext,omitempty"`
```
chris-sun-star, line 45: OperationContext is not used in controller

chris-sun-star (commented):
There's other scenarios needs to modify resource status other than delete and reset, maybe we need to support modify the operation context also, if not in a direct way, we can provides more types like resume, skip and retry for the case user has fixed the issue and want to continue the task

api/v1alpha1/obresourcerescue_types.go
PowerfooI: OK, I have removed it and update manifests with `make generate` and `make manifests`

internal/controller/obresourcerescue_controller.go
```diff
@@ -115,6 +115,43 @@ func (r *OBResourceRescueReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			idx, exist, err := unstructured.NestedInt64(uns.Object, "status", "operationContext", "idx")
+			if err != nil {
+				logger.Error(err, "failed to get idx of the target resource")
+				return ctrl.Result{}, err
+			}
+			if !exist {
+				logger.Error(err, "operationContext.idx not exist")
+				return ctrl.Result{}, nil
+			}
+			err = unstructured.SetNestedField(uns.Object, idx+1, "status", "operationContext", "idx")
```
chris-sun-star, line 145: what if idx points to the last task
also, TaskStatus needs to be set to Pending

chris-sun-star (approved):
LGTM

PR #152, opened 2024-01-02
fix #141: change failure rule of recover observer to retry from current

### LINKED ISSUE #141: [Bug]: create pod when really need
### Describe the bug

validate pod exists and status when recreate observer's pod, in some cases, recover task flow will break after pod recreated, or pod already recreated after K8s recovery

### Environment

A 3 node obcluster, pods recreated and observer stucks in recreate task flow

### Fast reproduce steps

1. create a multi-node obcluster
2. delete the pod
3. wait until pod recreation and observer not ready
4. kill ob-operator

### Expected behavior

the task flow continues to the end

### Actual behavior

task flow trys to recreate the code and always fails

### Additional context

_No response_

chris-sun-star:
## Summary

change strategy of failure rule to retry from current when recover observer
delete some debug related configurations under config

## Solution Description

After checking the code, first task of recover observer flow is to create the pod, the only operation that takes effect is to call the client to create the pod, simply change the retry strategy to from current task will prevent ob-operator trying to recreate the pod when it's already created
Tested with the following step:
1. create an obcluster
2. delete a pod
3. delete ob-operator's pod when observer is still recovering, so the task will fail
4. wait ob-operator restarts
5. check observer recovers

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -38,4 +38,4 @@ spec:
         - "--metrics-bind-address=127.0.0.1:8080"
         - "--leader-elect"
         - "--manager-namespace=oceanbase-system"
-        - "--log-verbosity=2"
+        - "--log-verbosity=0"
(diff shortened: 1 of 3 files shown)


PowerfooI:
LGTM

PowerfooI (approved):

PR #161, opened 2024-01-04
Experiments/standby

PowerfooI:
## Summary

1. Forbid to create empty standby tenant targeting at tenant in standalone cluster
2. Remove unused fields that telemetry may send
3. Move missing common type `observerTemplate` from api/v1alpha1 to api/types
4. Create default `standbyro` secret if users do not specify it in spec of tenant


chris-sun-star (approved):
LGTM

Issue #162, opened 2024-01-05
[Bug]: OBServer pod can not write volumes on OpenShift due to insufficient permissions


PowerfooI:
### Describe the bug

Related to #159
If the k8s distribution has some safety restriction, the observer using default service account may have no privilege to `mkdir` or `ln` in mounted volumes. However loosing restriction on default service account impacts other pods using the same default service account.

### Environment

OpenShift 4.14.6 hosted on AWS
With efs or gp3 as storage class

### Fast reproduce steps

1. Install ob-operator
2. Apply an obcluster.yaml 

### Expected behavior

The applied obcluster bootstrapped successfully.

### Actual behavior

The observer container throws "Prepare observer dir failed" error log and exit.

### Additional context

The `runAsUser` field of observer container is `1001030000`

Issue #159, opened 2024-01-03
[Question]: OceanBase K8S cluster setup


dimidevops:
### Description

Hi,

I'm trying to setup OB cluster on OpenShift 4.14.6 hosted on AWS.

I've installed the operator via helm (oceanbase/ob-operator:2.1.1) with no issues but when I try to deploy a test cluster I keep getting "PodSchedule" errors on the initial bootstrap pod.

![image](https://github.com/oceanbase/ob-operator/assets/147745543/bd32d6d7-3a9b-4e6f-9abc-1786bd7e5daa)

The yml I'm trying with:

```
apiVersion: oceanbase.oceanbase.com/v1alpha1
kind: OBCluster
metadata:
  name: seta-obcluster
  namespace: oceanbase-system
spec:
  clusterName: seta
  clusterId: 1
  userSecrets:
    root: root-password
    proxyro: proxyro-password
    monitor: monitor-password
    operator: operator-password
  topology:
    - zone: zone1
      replica: 1
      nodeSelector:
        kubernetes.io/hostname: ip-10-20-226-171.ec2.internal
  observer:
    image: 'oceanbase/oceanbase-cloud-native:4.2.1.1-101010012023111012'
    resource:
      cpu: 2
      memory: 10Gi
    storage:
      dataStorage:
        storageClass: efs
        size: 50Gi
      redoLogStorage:
        storageClass: efs
        size: 50Gi
      logStorage:
        storageClass: efs
        size: 20Gi
  monitor:
    image: 'oceanbase/obagent:4.2.1-100000092023101717'
    resource:
      cpu: 1
      memory: 1Gi
```

I'm trying to pin the pod to this worker node by using nodeSelector with kubernetes.io/hostname label, but it keeps saying "Insufficient memory" and as you can see the node has the available 11GB memory that is requested in the yml. Even when I don't specify the nodeSelector the same issue is present.

![image](https://github.com/oceanbase/ob-operator/assets/147745543/1b24558b-ddf2-48d8-b6f6-3ab9688b3add)

Am I doing something wrong or the nodes really cannot support the OB deployment or maybe this is some bug?

Also, I noticed that it refuses to be scheduled on nodes that are mixed as infra/worker.

Thank you.

PowerfooI:
Thank you for your feedback, first. @dimidevops 

The situation is awkward for us, too. I guess the cloud provider give your worker node less than 15.24 GiB memory that it shows. 

[...]

PR #167, opened 2024-01-05
Create specific service account for every single OBCluster

PowerfooI:
## Summary

1. Create specific service account named `<cluster-name>-obsa` and bind it to all observer pods
2. Before bootstrapping the obcluster, check and create the service account
3. Fix a logic mistake by the way at `observer_task.go:170`

about #159 #162


internal/const/oceanbase/oceanbase.go
```diff
@@ -171,3 +171,7 @@ const (
 const (
 	TolerateServerPodNotReadyMinutes = 5
 )
+
+const (
```
chris-sun-star, line 175: this value is useless

internal/controller/obcluster_controller.go
```diff
@@ -49,6 +49,7 @@ type OBClusterReconciler struct {
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=services/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups="",resources=services/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;update;patch;delete
```
chris-sun-star, line 52: also delete this

chris-sun-star (approved):
LGTM

PR #168, opened 2024-01-05
Remove httpOnly cookie on server

PowerfooI:
To delete an HttpOnly cookie, you must do so from the server side, since HttpOnly cookies are not accessible via JavaScript for security reasons. To delete such a cookie, the server must send an HTTP response with a Set-Cookie header that contains the same cookie name but with an expiration date in the past. This tells the client's browser to remove the cookie immediately.


chris-sun-star (approved):
LGTM

PR #174, opened 2024-01-15
fix:logo style

yang1666204:
## Summary
Fixed the problem of logo overflow pop-up window caused by style.
before

image

after

image


PowerfooI (approved):
LGTM

PR #175, opened 2024-01-17
Fix mismatched bool parameter due to mismatched cases

PowerfooI:
## Summary

In OceanBase database, bool parameters have value `True` or `False`. If we send `true` or `false`, they will never match.

## Solution Description

Compare desired value and actual value in lower case.


chris-sun-star (approved):
LGTM

PR #189, opened 2024-01-26
Refactor error handling logic; Add generic client for CRD; Add ctx as parameter for all inner handlers

PowerfooI:
## Summary

1. Create clients for every CRD type with the help of generics. They simplify defining new business logic greatly.
2. Wrap inner handlers with a  function ErrorWrapper (abbr. W), which handles all errors in it. So that we do not need SendResponse manually at all.
3. Add ctx as parameter for all inner handlers
4. Fix some tiny bugs like sending response more than once
5. Implement backup-related APIs


distribution/dashboard/pkg/errors/error.go
```diff
@@ -0,0 +1,68 @@
+package errors
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type oberror struct {
```
chris-sun-star, line 8: The error is actually HTTP status, call it oberror is misleading

distribution/dashboard/internal/handler/wrapper.go
```diff
@@ -0,0 +1,34 @@
...
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+	"github.com/oceanbase/oceanbase-dashboard/internal/model/response"
+	"github.com/oceanbase/oceanbase-dashboard/pkg/errors"
+)
+
+type Handler[T any] func(c *gin.Context) (T, error)
+
+func W[T any](h Handler[T]) gin.HandlerFunc {
```
chris-sun-star, line 13: Maybe Wrap is better, W is too short

distribution/dashboard/internal/handler/metric_handler.go
```diff
@@ -22,23 +24,21 @@ import (
...
-			logHandlerError(c, err)
-			SendInternalServerErrorResponse(c, nil, err)
-		} else {
-			SendSuccessfulResponse(c, metricClasses)
-		}
+		return nil, oberr.NewBadRequest(err.Error())
 	}
+	metricClasses, err := metric.ListMetricClasses(scope, language)
+	if err != nil {
+		logHandlerError(c, err)
```
chris-sun-star, line 38: logHandlerError can be moved into wrapper, thus we can log all the errors in one place, all the handlers only need to return the error

distribution/dashboard/pkg/errors/error.go
PowerfooI: OK, it's true. I have renamed it to `buzerror`.

distribution/dashboard/internal/handler/metric_handler.go
PowerfooI: Good idea! I have removed all loggings in specific handlers and moved error logging logic into the handler wrapper.

chris-sun-star (approved):

PR #192, opened 2024-01-29
Tenant overview

yang1666204:
## Summary 
Developing components such as Replicas,Events,and Backups,along with the necessary mock data for them.
image

### DIFF
--- a/distribution/dashboard/ui/src/pages/Tenant/Detail/Overview/Backups.tsx
+++ b/distribution/dashboard/ui/src/pages/Tenant/Detail/Overview/Backups.tsx
@@ -1,0 +1,101 @@
+import { ProCard } from '@ant-design/pro-components';
+import { Col, Descriptions, Table } from 'antd';
+import type { ColumnsType } from 'antd/es/table';
+
+interface BackupsProps {
+  backupPolicy: API.BackupPolicy;
+  backupJobs: API.BackupJob[];
+}
+
+export default function Backups({ backupPolicy, backupJobs }: BackupsProps) {
+  const PolicyConfig = {
+    destType: 'destType',
+    archivePath: '档案路径',
+    bakDataPath: 'bakDataPath',
+    scheduleType: '计划类型',
+    scheduleDates: '计划日期',
+    scheduleTime: '计划时间',
+  };
+
+  const columns: ColumnsType<API.BackupJob> = [
+    {
+      title: 'name',
+      dataIndex: 'name',
+      key: 'name',
+      width: 120,
+    },
+    {
+      title: '状态',
+      dataIndex: 'status',
+      key: 'status',
+      width: 120,
+    },
+    {
+      title: '数据库中状态',
+      dataIndex: 'statusInDatabase',
+      key: 'statusInDatabase',
+      width: 120,
+    },
+    {
+      title: '类型',
+      dataIndex: 'string',
+      key: 'string',
+      width: 120,
+    },
+    {
+      title: 'path',
+      dataIndex: 'path',
+      key: 'path',
+      width: 120,
+    },
+    {
+      title: 'encryptionSecret',
+      dataIndex: 'encryptionSecret',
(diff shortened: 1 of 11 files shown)


daolanx (dismissed):
+1

distribution/dashboard/ui/src/services/tenant.ts
```diff
@@ -98,8 +104,65 @@ export async function DeletePolicyOfTenant({
   });
 }
 
+export async function getBackupPolicy({
+  ns,
+  name,
+}: API.NamespaceAndName): Promise<API.BackupPolicyResponse> {
+  let r = await request(`${tenantPrefix}/${ns}/${name}/backupPolicy`);
+  let res: API.BackupPolicy = {
```
HzhOvO, line 112: Is it more intuitive and concise to simply retrieve values from the data based on the key?
const resKey = ['key1', 'key2'];
if (resKey.successful) {
  return {
    ...resKey,
    data: resKey.reduce((pre, cur) => {
      pre[key] = res.data?.[key];
      return pre
    }, pre)
  }
}

distribution/dashboard/ui/src/pages/Tenant/Detail/Overview/Backups.tsx
```diff
@@ -0,0 +1,101 @@
...
+          {Object.keys(PolicyConfig).map((key, index) => (
+            <Descriptions.Item label={PolicyConfig[key]} key={index}>
+              {key !== 'scheduleDates' ? (
+                <span>{backupPolicy[key]}</span>
+              ) : (
+                <div>
+                  {backupPolicy[key].map((item, index) => (
+                    <span key={index}>
+                      {item.backupType},{item.day}
+                      {index !== backupPolicy[key].length}{' '}
```
HzhOvO, line 84: {' '}
Does it have any practical effect, or is it just automatically inserted after formatting?

distribution/dashboard/ui/src/services/tenant.ts
```diff
@@ -98,8 +104,65 @@ export async function DeletePolicyOfTenant({
...
+    };
+  }
+  return r;
+}
+
+export async function getBackupJobs({
+  ns,
+  name,
+  type,
+  limit,
```
HzhOvO, line 136: limit = 10
Directly assigning default values to formal parameters would be more intuitive

distribution/dashboard/ui/src/services/tenant.ts
```diff
@@ -98,8 +104,65 @@ export async function DeletePolicyOfTenant({
...
+}: API.NamespaceAndName & {
+  type: API.JobType;
+  limit?: number;
+}): Promise<API.BackupJobsResponse> {
+  let r = await request(
+    `${tenantPrefix}/${ns}/${name}/backup/${type}/jobs?limit=${limit || 10}`,
```
HzhOvO, line 142: limit || 10
The default value should not be written here

distribution/dashboard/ui/src/services/tenant.ts
```diff
@@ -18,35 +18,41 @@ export async function getTenant({
...
-      clusterName:'',
-      tenantName:'',
-      tenantRole:'',
-      unitNumber:0,
-      status:'',
-      name:'',
-      namespace:'',
-      locality:''
+  });
+  let res: API.TenantBasicInfo = {
```
HzhOvO, line 22: Similarly to the following, I actually don't quite understand why objects are used here, nor is it necessary to set default values because they will be overwritten by forEach below. You can take a look at how to design them better

yang1666204: I need to use Object.keys to access the attributes of the info object.

distribution/dashboard/ui/src/services/tenant.ts
yang1666204: get

HzhOvO (approved):

PR #201, opened 2024-02-05
Example of sysbench

PowerfooI:
## Summary

1. Reduced value of reserved memory for possible overhead to exactly `90%` of pod resource limit
2. Added example of sysbench
3. Added cpu limit for containers in example `webapp`


PowerfooI:
By the way, I have searched [forum](https://ask.oceanbase.com) about `memory_limit`. It did not show that there are cases in which OceanBase consumes more memory than `memory_limit`. Maybe we could loose the `90%` limit to make OceanBase database gets more memory by default.

chris-sun-star (approved):
Maybe we can change the resource config under sysbench folder later, giving a production level configuration

PR #204, opened 2024-02-07
deploy ob-operator using terraform

chris-sun-star:
## Summary

Deploy ob-operator using terraform

### DIFF
--- a/README-CN.md
+++ b/README-CN.md
@@ -48,6 +48,29 @@ helm repo update
 helm install ob-operator ob-operator/ob-operator --namespace=oceanbase-system --create-namespace --version=2.1.2
 ```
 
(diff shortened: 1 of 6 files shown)


PowerfooI:
If there are more than one contexts in local `.kube/config`, `generate_k8s_cluster_tfvars.sh` will generate duplicated `tfvars`  which causes `Error: Attribute redefined` when executing `terraform apply`.

Adding `--minify`  to `kubectl config view` will solve it. 
```shell
$ kubectl config view --help
    --raw=false:
        Display raw byte data and sensitive data
```

The command will seem like below

```shell

```

PowerfooI (approved):
LGTM

PR #208, opened 2024-02-22
fix:Monitoring on the cluster overview page should be displayed in cluster dimensions

yang1666204:
## Summary
Modify the parameters of the monitoring query request to get names of clusters.

image


daolanx (approved):
+1

PR #213, opened 2024-02-28
feat:Modify backup policy

yang1666204:
## Summary

1. Add a page to modify the backup policy.
2. Separate some components from the new backup page to be shared by the modified backup page, such as: AdvancedConfiguration, BakMethodsList and ScheduleSelectComp, etc.

## Description
The page for modifying the backup policy has two states, editable state and display state.
image
image
And after initiating a request to modify or delete a backup, I couldn't get the latest backup policy immediately because it was asynchronous, so I did a poll for it.


yezaoshu (approved):
+1

PR #218, opened 2024-03-04
Fix bugs found in testings for OBResourceRescue

PowerfooI:
## Summary

1. Bare `0` could not be set in unstructured type. Set an int64 zero instead.
2. Refined logic of task manager by closing task result channel from sender end.


pkg/task/task_manager.go
```diff
@@ -159,11 +163,10 @@ func (m *TaskManager) CleanTaskResult(taskId tasktypes.TaskID) error {
 	if !exists {
 		return nil
 	}
-	retCh, ok := retChAny.(chan *tasktypes.TaskResult)
+	_, ok := retChAny.(chan *tasktypes.TaskResult)
```
chris-sun-star, line 166: this line is uncessary

chris-sun-star (approved):

PR #220, opened 2024-03-05
fix:Optimize UI

yang1666204:
## Summary

1. Unify the card style and Card instead of ProCard.
like this
before:
image
after:
image

3. Modify the fields jobKeepWindow, pieceInterval, and recoveryWindow to jobKeepDays, pieceIntervalDays, and recoveryDays.
4. Organize job list.
5. i18n.


yezaoshu (approved):
+1

PR #222, opened 2024-03-05
fix(statistic): add tenant statistic and update cluster statistic

PowerfooI:
## Summary

1. Added status statistic API for tenants.
2. Updated status statistic API of clusters which now returns all four status even though the count is `0`.

### DIFF
--- a/internal/dashboard/business/oceanbase/obcluster.go
+++ b/internal/dashboard/business/oceanbase/obcluster.go
@@ -546,8 +546,8 @@ func DeleteOBCluster(ctx context.Context, obclusterIdentity *param.K8sObjectIden
 	return oceanbase.DeleteOBCluster(ctx, obclusterIdentity.Namespace, obclusterIdentity.Name)
 }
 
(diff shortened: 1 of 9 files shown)


internal/dashboard/business/oceanbase/obcluster.go
```diff
@@ -543,28 +546,43 @@ func DeleteOBCluster(ctx context.Context, obclusterIdentity *param.K8sObjectIden
 	return oceanbase.DeleteOBCluster(ctx, obclusterIdentity.Namespace, obclusterIdentity.Name)
 }
 
-func GetOBClusterStatistic(ctx context.Context) ([]response.OBClusterStastistic, error) {
-	statisticResult := make([]response.OBClusterStastistic, 0)
+func GetOBClusterStatistic(ctx context.Context) ([]*response.OBClusterStastistic, error) {
```
chris-sun-star, line 549: Why change the return type to a list of pointers

PowerfooI: No special reason. Merely unify the style the same as obtenants' statistic API.

chris-sun-star (approved):

PR #224, opened 2024-03-06
Mount backup volume after standalone or service cluster starts

### LINKED ISSUE #196: [Enhancement]: Support for configuring backup volumes after a cluster starts.
### Description

It's forbidden to configure backup volumes after a cluster starts due to unrecoverable podIP. With clusterIP of service, maybe configuring backup volumes dynamically could be supported. Only a really short downtime would be introduced.

PowerfooI:
## Summary

1. Add support for mounting backup volume after cluster starts. Fixes #196.

### DIFF
--- a/api/v1alpha1/obcluster_webhook.go
+++ b/api/v1alpha1/obcluster_webhook.go
@@ -150,12 +150,26 @@ func (r *OBCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
 	if r.Spec.BackupVolume == nil && oldCluster.Spec.BackupVolume != nil {
 		return nil, errors.New("forbid to remove backup volume")
 	}
-	if r.Spec.BackupVolume != nil && oldCluster.Spec.BackupVolume == nil &&
-		mode != oceanbaseconst.ModeStandalone &&
-		mode != oceanbaseconst.ModeService {
-		return nil, errors.New("forbid to add backup volume on non-service or non-standalone cluster")
-	}
 	var err error
+	if r.Spec.BackupVolume != nil && oldCluster.Spec.BackupVolume == nil {
+		if mode != oceanbaseconst.ModeStandalone && mode != oceanbaseconst.ModeService {
+			observerList := &OBServerList{}
+			err = clt.List(context.TODO(), observerList)
+			if err != nil {
+				return nil, err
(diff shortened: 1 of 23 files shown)


api/v1alpha1/obcluster_webhook.go
```diff
@@ -147,6 +147,14 @@ func (r *OBCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
 	} else if oldMode != oceanbaseconst.ModeStandalone && (oldCluster.Spec.OBServerTemplate.Resource.Cpu != r.Spec.OBServerTemplate.Resource.Cpu || oldCluster.Spec.OBServerTemplate.Resource.Memory != r.Spec.OBServerTemplate.Resource.Memory) {
 		return nil, errors.New("forbid to modify cpu or memory quota of non-standalone cluster")
 	}
+	if r.Spec.BackupVolume == nil && oldCluster.Spec.BackupVolume != nil {
+		return nil, errors.New("forbid to remove backup volume")
+	}
+	if r.Spec.BackupVolume != nil && oldCluster.Spec.BackupVolume == nil &&
+		mode != oceanbaseconst.ModeStandalone &&
```
chris-sun-star, line 154: Is it possible to mount in normal mode, I think it only need to modify the container, the pod still remains

chris-sun-star (commented):
Is it possible to mount the volume without recreating the pod, it's only related to observer pod, is it possible to only recreate the container

api/v1alpha1/obcluster_webhook.go
chris-sun-star: It seems impossible to mount volume without recreating the pod after research, but we can still keep the ip using certain cni, which should be supported

PowerfooI: No, it's impossible to mount a new volume without modifying the pod. The `volumeMounts` field is in container but the declarations of `volumes` are in the spec of pod.

And by the way, if the service mode is available I don't think it is a good choice to deploy a "normal" mode cluster. :)

PowerfooI: OK, I'll check if the CNI is adopted.

chris-sun-star (approved):

PR #225, opened 2024-03-06
fix:i18n

yang1666204:
## Summary

1. Fixed inaccurate and incomplete nationalization issues.
2. Control the display or hiding of tenant recovery modules.


yezaoshu (approved):
+1

PR #246, opened 2024-03-14
Optimize dashboard api handlers

PowerfooI:
## Summary

1. Listing OBCluster returns brief responses now
2. Validated requests when deleting resource pool and creating backup policy
3. Change resource fields of int type to string for better rendering effects

### DIFF
--- a/internal/dashboard/business/oceanbase/obcluster.go
+++ b/internal/dashboard/business/oceanbase/obcluster.go
@@ -64,12 +64,12 @@ func getStatisticStatus(obcluster *v1alpha1.OBCluster) string {
 	}
 }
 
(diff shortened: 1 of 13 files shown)


internal/dashboard/business/oceanbase/obcluster.go
```diff
@@ -64,7 +64,100 @@ func getStatisticStatus(obcluster *v1alpha1.OBCluster) string {
 	}
 }
 
+func buildOBClusterBrief(ctx context.Context, obcluster *v1alpha1.OBCluster) (*response.OBClusterBrief, error) {
```
chris-sun-star, line 67: Maybe use overview instead of brief, it an adjective

chris-sun-star (approved):

PR #247, opened 2024-03-15
feat:Add events for cluster detail page

yang1666204:
## Summary
1. Add `EventsTable` for cluster detail page.
2. Add parameter `name` to the `EventsTable` component to filter events.
3. Add delete backup prompt.


yezaoshu (approved):
+1

PR #251, opened 2024-03-19
Some optimizations for dashboard handlers

powerfooI:
## Summary

1. Modified timestamp fields of `float64` type to `int64`
2. Renamed package `internal/oceanbase` to `internal/clients` for concision
3. Added events fields for backup policy that contains events of itself and its sub jobs
4. Added logging outputs for all gin handlers
5. Removed empty README in `ui/` folder
6. Fixed metrics plots with different amount of points by interpolation

### DIFF
--- a/internal/dashboard/router/v1/obcluster_router.go
+++ b/internal/dashboard/router/v1/obcluster_router.go
@@ -28,6 +28,5 @@ func InitOBClusterRoutes(g *gin.RouterGroup) {
 	g.POST("/obclusters/namespace/:namespace/name/:name/obzones", h.Wrap(h.AddOBZone))
 	g.POST("/obclusters/namespace/:namespace/name/:name/obzones/:obzoneName/scale", h.Wrap(h.ScaleOBServer))
 	g.DELETE("/obclusters/namespace/:namespace/name/:name/obzones/:obzoneName", h.Wrap(h.DeleteOBZone))
-	g.GET("/obclusters/:namespace/:name/essential-parameters", h.Wrap(h.ListOBClusterResources))
 	g.GET("/obclusters/:namespace/:name/resource-usages", h.Wrap(h.ListOBClusterResources))
 }


internal/dashboard/business/metric/metric.go
```diff
@@ -104,13 +105,35 @@ func extractMetricData(name string, resp *external.PrometheusQueryRangeResponse)
...
+		}
+		lenValues := len(values)
+		if lenValues == 0 {
+			continue
+		}
+		for i := range values {
+			// interpolate zero slot with average of previous and next value
+			if values[i].Value == 0 {
+				switch i {
+				case 0:
```
chris-sun-star, line 126: is it possible to solve the problem in front end, it's really normal to lose some datapoint

internal/dashboard/business/metric/metric.go
```diff
@@ -104,13 +105,35 @@ func extractMetricData(name string, resp *external.PrometheusQueryRangeResponse)
...
+					if lenValues > 1 {
+						values[i].Value = values[i+1].Value
+					}
+				case lenValues - 1:
+					if lenValues > 1 {
+						values[i].Value = values[i-1].Value
+					}
+				default:
```
chris-sun-star, line 134: what if more than one data point is lost, values[i+1] is 0, then values[i] equals values[i-1].Value / 2

internal/dashboard/router/v1/obcluster_router.go
```diff
@@ -29,4 +29,5 @@ func InitOBClusterRoutes(g *gin.RouterGroup) {
 	g.POST("/obclusters/namespace/:namespace/name/:name/obzones/:obzoneName/scale", h.Wrap(h.ScaleOBServer))
 	g.DELETE("/obclusters/namespace/:namespace/name/:name/obzones/:obzoneName", h.Wrap(h.DeleteOBZone))
 	g.GET("/obclusters/:namespace/:name/essential-parameters", h.Wrap(h.ListOBClusterResources))
```
chris-sun-star, line 31: don't forge to delete this line after front end modifies the url

PowerfooI: Oh yes, but the change need to be noticed by frontend app.

internal/dashboard/business/metric/metric.go
PowerfooI: I think it doesn't matter. It's a tradeoff which is better than hollow NaNs. Interpolation has a better rendering effects.

internal/dashboard/business/metric/metric.go
PowerfooI: No, I asked Yangon. The component he used in frontend can not handle the NaN. Interpolation is an inevitable choice.

internal/dashboard/business/metric/metric.go
PowerfooI: I refined the interpolation logic and now it concerns continuous NaNs.

chris-sun-star (approved):

PR #255, opened 2024-03-20
fix:Filter monitored parameters

yang1666204:
## Summary
1. Filter out tenants or clusters that no longer exist from monitoring data.
image

3. Delete modal uniformly.
image

1. Filter out the failed cluster


huhansan3 (dismissed):
+1

huhansan3 (approved):
+2

PR #256, opened 2024-03-20
perf:Avoid the monitoring module repeatedly requesting clusters or tenants

yang1666204:
## Summary
1. Separate the ability to obtain clusters or tenants from monitoring requests to avoid a large number of requests.
2. Some UI bugfix.


yezaoshu (approved):
+1

PR #257, opened 2024-03-20
Fix create tenant

yang1666204:
## Summary

1. Add fields for tenant detail page.
2. Creating a tenant allows the resource pool to have no weight filled in.
3. Add `IconTip` component from ob-cloud.


yezaoshu (approved):
+1

PR #266, opened 2024-03-27
fix:Supplement available resources in the resource pool

yang1666204:
## Summary

1. Supplement available resources in the resource pool when editing the resource pool.
2. Add `version` number.
3. Optimize the topology map with more operation lists and add the attribute `disabled` for each item.


yezaoshu (approved):
+1

PR #268, opened 2024-03-27
add example for deloy obconfigserver using mysql as backend

chris-sun-star:
## Summary

add example config files for deploying ob-configserver using mysql as backend


PowerfooI (approved):
LGTM

PR #274, opened 2024-03-29
Optimize some methods of dashboard

powerfooI:
## Summary

1. Added UID for resource response and returned the exact type from handlers
2. Added RBAC permission for dashboard to discover `ob-operator` deployment
3. Fixed some typos and logic redundancy; generated documents
4. Sorted results of listing OBCluster, OBZone and OBServer eliminating randomness due to k8s list
5. Restricted OBCluster creation if there is obviously no enough resource for the pods


internal/dashboard/business/oceanbase/obcluster.go
```diff
@@ -478,64 +492,125 @@ func generateOBClusterInstance(param *param.CreateOBClusterParam) *v1alpha1.OBCl
...
+func ScaleOBServer(ctx context.Context, obzoneIdentity *param.OBZoneIdentity, scaleParam *param.ScaleOBServerParam) (*response.OBCluster, error) {
 	obcluster, err := clients.GetOBCluster(ctx, obzoneIdentity.Namespace, obzoneIdentity.Name)
 	if err != nil {
-		return errors.Wrapf(err, "Get obcluster %s %s", obzoneIdentity.Namespace, obzoneIdentity.Name)
-	}
-	if obcluster.Status.Status != clusterstatus.Running {
-		return errors.Errorf("Obcluster status invalid %s", obcluster.Status.Status)
+		return nil, errors.Wrapf(err, "Get obcluster %s %s", obzoneIdentity.Namespace, obzoneIdentity.Name)
 	}
+	// if obcluster.Status.Status != clusterstatus.Running {
```
chris-sun-star, line 567: uncomment this

internal/dashboard/business/oceanbase/obtenant.go
```diff
@@ -244,7 +244,7 @@ func CreateOBTenant(ctx context.Context, nn types.NamespacedName, p *param.Creat
 			Namespace: nn.Namespace,
 		},
 		StringData: map[string]string{
-			"password": rand.String(20),
+			"password": p.RootPassword, // For simplicity, use the same password as root
```
chris-sun-star, line 247: it's not a good practice to use the same password

internal/dashboard/handler/info_handler.go
```diff
@@ -134,11 +163,32 @@ func GetStatistics(c *gin.Context) (*response.StatisticDataResponse, error) {
 	if err != nil {
 		return nil, httpErr.NewInternal(err.Error())
 	}
+
+	// Get deployment named oceanbase-controller-manager in oceanbase-system namespace
+	deployment, err := clt.ClientSet.AppsV1().Deployments("oceanbase-system").Get(c, "oceanbase-controller-manager", metav1.GetOptions{})
```
chris-sun-star, line 168: we need to search all namespaces

chris-sun-star (commented):
is it possible to move resource validation into webhook, thus we don't need to reimpliment the same logic in dashboard

internal/dashboard/business/oceanbase/obcluster.go
```diff
@@ -478,64 +492,125 @@ func generateOBClusterInstance(param *param.CreateOBClusterParam) *v1alpha1.OBCl
 	return obcluster
 }
 
-func CreateOBCluster(ctx context.Context, param *param.CreateOBClusterParam) error {
+func JudgeResourceEnoughForOBCluster(ctx context.Context, obcluster *v1alpha1.OBCluster) error {
```
chris-sun-star, line 495: consider node selector and affinity?

internal/dashboard/business/oceanbase/obtenant.go
powerfooI: This is necessary if we require users to submit one password field.

powerfooI:
> is it possible to move resource validation into webhook, thus we don't need to reimpliment the same logic in dashboard

I think it's not reasonable to return error message thrown by webhook to dashboard users who do not even know the CRD structure. The error messages thrown from webhook are quite low level which describe the fields of underwater CRs.

internal/dashboard/handler/info_handler.go
powerfooI: Maybe we can use `List` method and take the first one in returned items. How do you think about it?

internal/dashboard/business/oceanbase/obtenant.go
powerfooI: Otherwise, it brings problems when restoring a tenant from backup data. Currently it would generate a different password from original.

internal/dashboard/business/oceanbase/obcluster.go
powerfooI: It's difficult to analyze affinity in, this function is merely a loose or an approximate way to judge whether all remain resource is enough for new coming oceanbase cluster. It prevents some mistakes.

chris-sun-star: I will suggest to delete the validation then, if you can't do the validation precisely by design, then it's better not to do it.

internal/dashboard/business/oceanbase/obtenant.go
chris-sun-star: If the user need to restore or create standby tenant, we can give them the way to retrieve this password, or we do it automatically, we can discuss later.

chris-sun-star:
> > is it possible to move resource validation into webhook, thus we don't need to reimpliment the same logic in dashboard
> 
> I don't think it's reasonable to return error message thrown by webhook to dashboard users who do not even know the CRD structure. The error messages thrown from webhook are quite low level which describe the fields of underwater CRs.

I just mean the validation logic can be reused, you can return a customized error message in dashboard

powerfooI:
> > > is it possible to move resource validation into webhook, thus we don't need to reimpliment the same logic in dashboard
> > 
> > 
> > I don't think it's reasonable to return error message thrown by webhook to dashboard users who do not even know the CRD structure. The error messages thrown from webhook are quite low level which describe the fields of underwater CRs.
> 
> I just mean the validation logic can be reused, you can return a customized error message in dashboard

It's true! Maybe we could extract complex CRD validation logic as an internal package. I'll do this in one of next few PRs.

internal/dashboard/business/oceanbase/obcluster.go
powerfooI: Let's discuss about it tomorrow ~

PR #275, opened 2024-03-29
Feat statistics

yang1666204:
## Summary
1. Regular statistical data after entering the system.
2. Separate the parameter `Version` from the operation and maintenance operation requests so that statistics reporting is not aware of when calling these requests.

### DIFF
--- a/ui/src/utils/helper.ts
+++ b/ui/src/utils/helper.ts
@@ -1,4 +1,7 @@
 import type { PoolDetailType } from '@/components/customModal/ModifyUnitDetailModal';
+import { getAppInfo, getStatistics } from '@/services';
+import { REPORT_PARAMS_MAP, reportData } from '@/services/reportRequest';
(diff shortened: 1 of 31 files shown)


ui/src/utils/helper.ts
```diff
@@ -66,3 +72,35 @@ export const formatPatchPoolData = (originUnitData: PoolDetailType,type:'edit'|'
   }
   return newOriginUnitData;
 };
+
+export const getAppInfoFromStorage = async (): Promise<API.AppInfo> => {
```
powerfooI, line 76: What if the version changes?

yang1666204: SessionStorage will be cleared when the window is closed or the browser is closed. Do users need to build different packages when using different versions? If it is, there will be no impact

powerfooI: I see, storing in sessionStorage is OK.

powerfooI (approved):
LGTM

PR #276, opened 2024-04-01
Trim string

yang1666204:
## Summary
1. Clear spaces at the beginning and end of a string.
2. Limit the number of zones in the cluster to less than or equal to 2 and cannot be deleted.


powerfooI (approved):
LGTM

PR #279, opened 2024-04-01
Fix resource pool related defects

yang1666204:
## Summary
1. Restrict the tenant's Topo graph to prevent the resource pool from being operated when the tenant's status or the cluster's status is not running.
2. Fix the problem that publicKey is not updated because it is stored in sessionStorage.
3. Fix issues related to resource pools.

### DIFF
--- a/ui/src/hook/usePublicKey.ts
+++ b/ui/src/hook/usePublicKey.ts
@@ -1,18 +1,21 @@
 import { getAppInfo } from '@/services';
+import { useModel } from '@umijs/max';
 import JSEncrypt from 'jsencrypt';
-import React, { useEffect } from 'react';
+import { useEffect } from 'react';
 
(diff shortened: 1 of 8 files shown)


ui/src/hook/usePublicKey.ts
```diff
@@ -1,23 +1,28 @@
...
   const [publicKey, setPublicKey] = React.useState<string>('');
 
   useEffect(() => {
-    getAppInfoFromStorage().then((appInfo) => {
-      setPublicKey(appInfo.publicKey); 
-    }).catch((err) => {
-      console.log(err)
-    });
-  }, [])
+    getAppInfo()
```
powerfooI, line 9: Had better keep the public key in memory instead of send request every time.

powerfooI (approved):

PR #282, opened 2024-04-02
Add loading status

yang1666204:
## Summary
1. Add loading status when making a request.
2. Fix the problem of cluster mode not mapping value.


powerfooI (approved):
LGTM

PR #283, opened 2024-04-03
fix:Refresh tooltip content

yang1666204:
## Summary
1. The topology map resource pool is updated to tooltip after editing.
2. Solve the problem of occasional white screen when editing resource pool.
3. The main tenant synchronization function in the created tenant filters out the running tenants.
4. Solve some UI issues.

### DIFF
--- a/ui/src/components/TopoComponent/helper.ts
+++ b/ui/src/components/TopoComponent/helper.ts
@@ -223,7 +223,7 @@ export const formatTopoData = (
 /**
  * Determine whether the old and new topoData attribute values are exactly the same
  */
-export const checkIsSame = (oldTopoData: any, newTopoData: any): boolean => {
+export const checkTopoDataIsSame = (oldTopoData: any, newTopoData: any): boolean => {
   if (!_.matches(oldTopoData)(newTopoData)) return false;
   if (newTopoData.children.length > oldTopoData.children.length) return false;
   oldTopoData.children.forEach((oldZone: any, idx: number) => {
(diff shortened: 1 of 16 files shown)


ui/src/pages/Tenant/New/TenantSource.tsx
```diff
@@ -1,25 +1,24 @@
...
-  DatePicker,
-  Form,
-  Input,
-  Row,
-  Select,
-  Space,
-  Switch,
-  TimePicker,
-  Tooltip,
+Card,
```
powerfooI, line 5: Why do you remove leading spaces here?

ui/src/components/TopoComponent/helper.ts
```diff
@@ -233,6 +233,22 @@ export const checkIsSame = (oldTopoData: any, newTopoData: any): boolean => {
   return true;
 };
 
+// export const checkReplicasIsSame = (
```
powerfooI, line 236: Remove this if it's unused.

powerfooI (changes requested):

ui/src/pages/Tenant/New/TenantSource.tsx
yang1666204: The internationalization tool automatically formats.

yang1666204: But I can modify the configuration of the tool so that it formats the code in the ideal way

powerfooI (approved):
LGTM

PR #286, opened 2024-04-07
fix:Computing resources

yang1666204:
## Summary
Correct the way resource amounts are calculated.

### DIFF
--- a/ui/src/pages/Tenant/helper.ts
+++ b/ui/src/pages/Tenant/helper.ts
@@ -197,11 +197,11 @@ export function findMinParameter(
         ),
         maxLogDisk: Math.max(
           result.maxLogDisk!,
-          essentialParameter[zone]['availableCPU'],
+          essentialParameter[zone]['availableLogDisk'],
         ),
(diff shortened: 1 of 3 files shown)


ui/src/pages/Tenant/helper.ts
```diff
@@ -175,29 +175,38 @@ export function checkIsSame(
...
+      };
+    } else {
+      result = {
+        maxCPU: Math.max(
+          result.maxCPU!,
+          essentialParameter[zone]['availableCPU'],
+        ),
+        maxLogDisk: Math.max(
+          result.maxLogDisk!,
+          essentialParameter[zone]['availableCPU'],
```
powerfooI, line 200: Wrong field here, you should access diskSize and memory.

powerfooI (changes requested):

ui/src/pages/Tenant/helper.ts
yang1666204: ok

powerfooI (approved):
LGTM

PR #290, opened 2024-04-09
Added timeout for infinitely looping task

powerfooI:
## Summary

1. If a user set a wrong target image when upgrading OBCluster, created job will never end or be deleted. This PR added a long timeout and OwnerReference for infinitely looping task to avoid memory leak.
2. Generated truly random passwords instead of hardcode plain password for 3 underwater users when creating OBCluster with dashboard.


chris-sun-star (commented):
Maybe it better to make the timeout logic a wrapper function

powerfooI:
> Maybe it better to make the timeout logic a wrapper function

OK, I have wrap the checking with timeout logic in a single function~

chris-sun-star (approved):

PR #294, opened 2024-04-10
Some usability optimizations

yang1666204:
## Summary
1. Resource pool data is rounded down to `two digits`.
image

3. Avoid invalid `loading` caused by no change in `tenantDetail` during polling.


powerfooI (approved):
LGTM

PR #295, opened 2024-04-10
Fixed deletion of resource pool parameters

yang1666204:
## Summary
The cluster resource name used to delete the resource pool in the topo diagram is corrected to use the tenant resource name.


powerfooI (approved):
LGTM

PR #335, opened 2024-04-25
Fixed inter-cluster empty standby

powerfooI:


chris-sun-star (dismissed):

internal/resource/utils/tenant.go
```diff
@@ -24,59 +24,97 @@ import (
...
+	for _, ls := range logStats {
+		if ls.BeginLSN != 0 {
+			return errors.New("Log stats begin SCN is not 0, log is of not integrity")
+		}
+	}
+
+	return nil
+}
+
+func getTenantAndClusterCon(ctx context.Context, clt client.Client, logger *logr.Logger, ns, tenantCR string) (*v1alpha1.OBTenant, *operation.OceanbaseOperationManager, error) {
```
chris-sun-star, line 86: split this into two function, get tenant and get sys connection
besides, the tenant may not be primary, so the parameter 'primary' can be changed to a more general name

chris-sun-star (approved):

Issue #345, opened 2024-04-26
[Bug]: dashboard only displays one zone but the selected obcluster actually has 3 zone


chris-sun-star:
### Describe the bug

![image](https://github.com/oceanbase/ob-operator/assets/85611200/ac0f4d8f-f8ba-481a-96bd-8561e035912c)
![image](https://github.com/oceanbase/ob-operator/assets/85611200/64fe9ccc-19cf-4af8-ae32-a284953f31bd)

dashboard only displays one zone but the selected obcluster actually has 3 zone

### Environment

oceanbase-dashboard 0.2.1
ob-operator 2.2.0

### Fast reproduce steps

create a two obcluster, one with one zone and the other has three zones
create tenant using oceanbase-dashboard, select the cluster with three zones as destination obcluster

PR #346, opened 2024-04-28
fix:The selected cluster exception display zone

yang1666204:
## Summary
Fix problems caused by duplicate clusterId  #345


ui/src/services/index.ts
```diff
@@ -155,18 +154,19 @@ export async function getSimpleClusterList(): Promise<API.SimpleClusterListRespo
...
   });
   if (r.successful) {
-    return{
+    return {
       ...r,
-      data:r.data.map((clusterDetail) => ({
-        clusterId: clusterDetail.clusterId,
+      data: r.data.map((clusterDetail) => ({
+        clusterId: clusterDetail.clusterId, // clusterId is not unique
+        id: clusterDetail.name + clusterDetail.namespace,
```
powerfooI, line 161: Use unambiguous id like `$namespace:$name`. Concatenating the name and namespace with `+` will lead to potential collision. e.g. resource named `ab` in namespace `cd` will has the same id with another resource named `a` in namespace `bcd`.

powerfooI (changes requested):

powerfooI (approved):
LGTM

PR #350, opened 2024-05-06
Add openapi generator api

yang1666204:
## Summary
Reference documentation：https://github.com/baurine/swag-example/blob/main/notes/note.md


ui/src/api/index.ts
```diff
@@ -0,0 +1,39 @@
+/* tslint:disable */
+/* eslint-disable */
+/**
```
powerfooI, line 3: This file is not generated, remove comments here

powerfooI (changes requested):

ui/src/api/index.ts
yang1666204: ok

powerfooI (approved):
LGTM

PR #358, opened 2024-05-09
Fixed an issue where tenant monitoring charts were displayed abnormally in cluster details

yang1666204:
## Summary

1. The tenant monitoring chart in the cluster details should only display the tenants currently existing under the cluster and the sys tenant.
image

2.Regenerate api.


ui/src/api/generated/base.ts
```diff
@@ -16,7 +16,7 @@
 import type { Configuration } from './configuration';
 // Some imports not used depending on template conditions
 // @ts-ignore
-import type { AxiosInstance, RawAxiosRequestConfig } from 'axios';
+import type { AxiosPromise, AxiosInstance, RawAxiosRequestConfig } from 'axios';
```
powerfooI, line 19: This `AxiosPromise` import is unused in this file. It's OK to remove it.

powerfooI (changes requested):

ui/src/api/generated/base.ts
yang1666204: In fact this line of code appears every time we regenerate, can we ignore it?

powerfooI: OK, just ignore it if it's generated.

powerfooI (approved):
LGTM

PR #361, opened 2024-05-09
Feat:add the drawer component of the alarm channel

yang1666204:
## Summary
As shown in the picture:
image


ui/src/pages/Alert/Channel/ChannelDrawer.tsx
```diff
@@ -0,0 +1,164 @@
...
+        </Space>
+      </div>
+    );
+  };
+
+  useEffect(() => {
+    if (status !== 'create' && name) {
+      getReceiver(name);
+    }
+  }, [status]);
```
powerfooI, line 69: Why not put the name in the dependency list? I thought it's a dependency in this effect, too.

ui/src/pages/Alert/Channel/ChannelDrawer.tsx
```diff
@@ -0,0 +1,164 @@
...
+    onSuccess: ({ successful, data }) => {
+      if (successful) {
+        form.setFieldsValue({
+          ...data,
+        });
+      }
+    },
+  });
+  const { data: listReceiversRes } = useRequest(alert.listReceivers, {
+    manual: true,
```
powerfooI, line 41: Why set `manual: true` here while no `run` used in following part? If you don't run manually, the `data` (or `listReceiversRes`) will always be null.

powerfooI (changes requested):

powerfooI (approved):
LGTM

PR #364, opened 2024-05-10
Fix:resource verification depends on zone selection

yang1666204:
## Summary
Fixed the issue where switching zone selection was not re-verified.
image
image


ui/src/pages/Tenant/helper.ts
```diff
@@ -179,6 +179,12 @@ export function checkIsSame(
...
 }
 
+/**
+ * The minimum value of available resources in the currently selected zone 
+ * is the maximum value that can be entered in the input box.
+ * 
+ * The reason is to ensure that the input resource specifications can be created by each zone
+ */
 export function findMinParameter(
   zones: string[],
```
powerfooI, line 189: Is it possible that `zones` is empty here? If so, the result would be a empty object and accessing its fields like `result.maxCPU` will cause problem.

powerfooI (approved):
LGTM

PR #393, opened 2024-05-21
Supports selecting shield objects when creating a shield

yang1666204:
## Summary

1. Supports selecting shield objects when creating a shield.
image

3. Upgrade the antd version to support the maxCount attribute of the Select component.
4. Support clearing form data after closing the drawer.


ui/src/pages/Alert/Shield/index.tsx
```diff
@@ -19,7 +19,7 @@ const { Text } = Typography;
 
 export default function Shield() {
   const [form] = Form.useForm();
-  const [searchParams] = useSearchParams();
+  const [searchParams,setSearchParams] = useSearchParams();
```
powerfooI, line 22: Had better add a space behind a comma.

ui/src/pages/Alert/index.tsx
```diff
@@ -35,8 +35,18 @@ const items: TabsProps['items'] = [
     label: '告警推送',
   },
 ];
+
+const getInitialActiveKey = () => {
+  const tailKey = location.hash.split('/')[location.hash.split('/').length - 1];
```
powerfooI, line 40: Why not declare a const like `const hashFrags = location.hash.split('/')`? Splitting the same string twice is not so good.

ui/src/pages/Alert/index.tsx
```diff
@@ -35,8 +35,18 @@ const items: TabsProps['items'] = [
...
+  if (
+    ['event', 'shield', 'rules', 'channel', 'subscriptions'].includes(tailKey)
```
powerfooI, line 42: Declare the array as a constant. If it does not change with the context, declare it outside the function.

powerfooI (approved):
LGTM

PR #403, opened 2024-05-23
Supports requesting data after modifying filter items

yang1666204:
## Summary
Supports requesting data after modifying filter items and add an anti-shake mechanism to the request function.


powerfooI (approved):
LGTM

PR #404, opened 2024-05-23
Restrict scaling down if new memory size is less than assigned memory

powerfooI:

### DIFF
--- a/api/v1alpha1/obcluster_webhook.go
+++ b/api/v1alpha1/obcluster_webhook.go
@@ -39,6 +39,7 @@ import (
 	apitypes "github.com/oceanbase/ob-operator/api/types"
 	obcfg "github.com/oceanbase/ob-operator/internal/config/operator"
 	oceanbaseconst "github.com/oceanbase/ob-operator/internal/const/oceanbase"
+	clusterstatus "github.com/oceanbase/ob-operator/internal/const/status/obcluster"
 )
 
 // log is for logging in this package.
@@ -49,8 +50,6 @@ func (r *OBCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		For(r).
 		Complete()
 }
-
-// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
 
 //+kubebuilder:webhook:path=/mutate-oceanbase-oceanbase-com-v1alpha1-obcluster,mutating=true,failurePolicy=fail,sideEffects=None,groups=oceanbase.oceanbase.com,resources=obclusters,verbs=create;update,versions=v1alpha1,name=mobcluster.kb.io,admissionReviewVersions=v1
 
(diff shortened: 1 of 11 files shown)


api/v1alpha1/obcluster_webhook.go
```diff
@@ -186,6 +185,42 @@ func (r *OBCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
...
+				maxAssignedMemory = gvserver.MemAssigned
+			}
+			if gvserver.CPUAssigned > maxAssignedCPU {
+				maxAssignedCPU = gvserver.CPUAssigned
+			}
+		}
+		if newResource.Memory.AsApproximateFloat64()*memoryLimitPercent < float64(maxAssignedMemory) {
+			return nil, errors.New("Assigned memory is larger than new memory size")
+		}
+		if newResource.Cpu.Value() > 16 && newResource.Cpu.AsApproximateFloat64() < float64(maxAssignedCPU) {
```
chris-sun-star, line 220: 16 is the default cpu count set to obcluster but not the actual resource, and this validations is under the condition of memory shrink, but cpu can extend, for example, the obcluster is created with cpu = 2, and the cpu count will be 16 in obcluster, then change it to 4, it will not pass your validation.

internal/resource/obcluster/obcluster_task.go
```diff
@@ -1112,6 +1112,97 @@ func AnnotateOBCluster(m *OBClusterManager) tasktypes.TaskError {
...
+			}
+			maxAssignedMem = gvserver.MemAssigned
+		}
+		if gvserver.CPUAssigned > maxAssignedCpu {
+			maxAssignedCpu = gvserver.CPUAssigned
+		}
+	}
+	newResource := m.OBCluster.Spec.OBServerTemplate.Resource
+	specMem := newResource.Memory.AsApproximateFloat64()
+	specWithPercent := int64(specMem * memoryLimitPercent)
```
chris-sun-star, line 1155: change the variable to specMemoryLimit

internal/resource/obcluster/obcluster_task.go
```diff
@@ -1182,7 +1182,7 @@ func AdjustParameters(m *OBClusterManager) tasktypes.TaskError {
 		}
 	}
 
-	if newResource.Cpu.Value() > 16 && oldResource.Cpu.Cmp(newResource.Cpu) != 0 {
+	if oldResource.Cpu.Value() > 16 && oldResource.Cpu.Cmp(newResource.Cpu) != 0 {
 		foundCpuCount := false
 		for i, p := range copiedCluster.Spec.Parameters {
 			if p.Name == "cpu_count" {
```
chris-sun-star, line 1188: calculate cpu count first, if newResource.Cpu is less than 16, set it to 16, and oldResource.Cpu.Value() > 16 is not necessary.

powerfooI: OK, you are right. I got confused here hahah.

chris-sun-star (approved):

PR #406, opened 2024-05-24
Feature: Modify configurations of observers w/o stopping running cluster

powerfooI:
## Summary

1. Rolling update zones and servers if server template is detected to differ from previous.
2. Trigger rolling update of zones one by one. In the scope of a zone, every single old server would be deleted safely (protected by database internal mechanism) only after a new server is added into the zone and runs as expected.


internal/const/status/obcluster/obcluster_status.go
```diff
@@ -13,18 +13,19 @@ See the Mulan PSL v2 for more details.
...
+	ModifyOBZoneReplica    = "modify obzone replica"
+	Upgrade                = "upgrade"
+	ModifyOBParameter      = "modify parameter"
+	Bootstrapped           = "bootstrapped"
+	FinalizerFinished      = "finalizer finished"
+	ScaleUp                = "scale up"
+	ExpandPVC              = "expand pvc"
+	Failed                 = "failed"
+	MountBackupVolume      = "mount backup volume"
+	RollingUpdateOBServers = "rolling update observers"
```
chris-sun-star, line 30: does obcluster update observers directly or just rolling update obzones

internal/resource/obzone/obzone_task.go
```diff
@@ -426,3 +369,46 @@ func WaitForOBServerExpandingPVC(m *OBZoneManager) tasktypes.TaskError {
 func WaitForOBServerMounting(m *OBZoneManager) tasktypes.TaskError {
 	return m.generateWaitOBServerStatusFunc(serverstatus.MountBackupVolume, obcfg.GetConfig().Time.DefaultStateWaitTimeout)()
 }
+
+func RollingUpdateOBServers(m *OBZoneManager) tasktypes.TaskError {
```
chris-sun-star, line 373: RollingReplaceOBServers may be a precise name

chris-sun-star (approved):

PR #409, opened 2024-05-28
Initialize proxy management module

yang1666204:
## Summary
As shown in the pictures:

image
image


ui/src/pages/OBProxy/ClusterList.tsx
```diff
@@ -0,0 +1,90 @@
...
+    dataIndex: 'Replicas',
+    key: 'Replicas',
+  },
+  {
+    title: 'serviceIp',
+    dataIndex: 'serviceIp',
+    key: 'serviceIp',
+  },
+  {
+    title: 'creationTime',
```
powerfooI, line 52: Use Chinese here. The same as `replicas`, `serviceIP` and `status` fields.

powerfooI (changes requested):

ui/src/pages/OBProxy/ClusterList.tsx
yang1666204: Done.

powerfooI (approved):
LGTM

PR #422, opened 2024-06-04
Helper image

chris-sun-star:
## Summary

split docker image of oceanbase-cloud-native into 2 parts
oceanbase-helper which is built from ob-operator's source code, build and publish image when ob-operator releases
oceanbase-cloud-native uses the latest oceanbase-helper image as base image


powerfooI (approved):
LGTM

PR #423, opened 2024-06-04
Check label input box

yang1666204:
## Summary

1. Dynamically validate the label input box.
2. Filter out empty labels when submitting a form.

### DIFF
--- a/ui/src/components/InputLabel/index.tsx
+++ b/ui/src/components/InputLabel/index.tsx
@@ -1,14 +1,32 @@
 import { DeleteOutlined, PlusOutlined } from '@ant-design/icons';
-import { Button, Checkbox, Col, Form, Input, Popconfirm, Row } from 'antd';
+import {
+  Button,
+  Checkbox,
+  Col,
+  Form,
+  FormInstance,
+  Input,
+  Popconfirm,
+  Row,
+} from 'antd';
+import { ReactElement } from 'react';
 import styles from './index.less';
 
 interface InputLabelProps {
   wrapFormName: string;
   labelFormName: string;
   valueFormName: string;
+  form: FormInstance<unknown>;
   regBoxFormName?: string;
   showDelete?: boolean;
   maxCount?: number;
+}
+
+interface LableFormItemProps {
+  name: (string | number)[];
+  fieldIdx: number;
+  dependName: (string | number)[];
+  children: ReactElement;
 }
 
 export default function InputLabel({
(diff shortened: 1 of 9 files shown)


yezaoshu (approved):
+1

powerfooI (changes requested):
Pending

ui/src/components/InputLabel/index.tsx
```diff
@@ -1,24 +1,81 @@
...
   regBoxFormName?: string;
   showDelete?: boolean;
   maxCount?: number;
 }
 
+interface LableFormItemProps {
+  name: (string | number)[];
+  fieldIdx: number;
+  dependName: (string | number)[];
+  children: ReactElement;
```
powerfooI, line 29: It's more common to use `ReactNode` as children

ui/src/components/InputLabel/index.tsx
```diff
@@ -27,26 +84,42 @@ export default function InputLabel({
             {fields.map(({ key, name }, index) => (
               <Row gutter={8} style={{ marginBottom: 8 }} key={key}>
                 <Col span={11}>
-                  <Form.Item name={[name, labelFormName]} noStyle>
+                  <LableFormItem
+                    name={[name, labelFormName]}
+                    fieldIdx={name}
+                    dependName={[name, valueFormName]}
```
powerfooI, line 90: Is this `valueFormName` confused with following form item? Why use `valueFormName` with `label` and `labelFormName` with `value`?

ui/src/pages/Alert/Shield/ShieldDrawerForm.tsx
```diff
@@ -71,9 +72,11 @@ export default function ShieldDrawerForm({
     } else {
       form.setFieldValue('endsAt', dayjs(time));
```
powerfooI, line 73: I found that there are dayjs and moment in the `package.json`. It would be a good choice to use only one of them.

ui/src/pages/Alert/Subscriptions/SubscripDrawerForm.tsx
```diff
@@ -102,6 +104,7 @@ export default function SubscripDrawerForm({
                 labelFormName="name"
                 valueFormName="value"
                 regBoxFormName="isRegex"
+                form={form}
```
powerfooI, line 107: This is not a good pattern. `Form.Item` is controlled by `Form`, you added dependency from item to form,  which construct a control cycle here.

powerfooI (changes requested):

ui/src/components/InputLabel/index.tsx
yang1666204: `valueFormName` means the form name of the value input box in the Label component.

yang1666204: The `label name` and `label value` verification are dependent on each other. When one of them is empty, the other can also be empty; when one of them is not empty, the other is not empty either.

ui/src/pages/Alert/Shield/ShieldDrawerForm.tsx
yang1666204: You are right, `dayjs` should be able to completely replace `moment`.

powerfooI: `dayjs` seems to be more light-weight and easy to use

yang1666204: And `antd` recommends using `dayjs`.

ui/src/pages/Alert/Subscriptions/SubscripDrawerForm.tsx
powerfooI: If you want to set field values in `form`, define a controlled form item instead. Reference of controlled form item could be found at https://ant-design.antgroup.com/components/form-cn#components-form-demo-customized-form-controls

yang1666204: Review again?

PR #430, opened 2024-06-06
Some details

yang1666204:
## Summary

1. Field modification: `serverity` ->`severity`
2. Fixed some UI issues.
3. Adjust `InputLabel` validation rules.


ui/src/pages/Alert/AlarmFilter/index.tsx
```diff
@@ -43,28 +43,60 @@ export default function AlarmFilter({ form, type, depend }: AlarmFilterProps) {
...
     if (type === 'obtenant') {
-      return (list as Alert.TenantsList[]).map((cluster) => ({
-        label: <span>{cluster.clusterName}</span>,
-        title: cluster.clusterName,
-        options: cluster.tenants?.map((item) => ({
-          value: item,
-          label: item,
-        })),
-      }));
+      return (list as Alert.TenantsList[])
```
powerfooI, line 46: Why declare type here instead of where the variable was initialized? Do not use so many `as`

ui/src/pages/Alert/Shield/index.tsx
```diff
@@ -70,6 +70,7 @@ export default function Shield() {
       title: '屏蔽应用/对象类型',
       dataIndex: 'instances',
       key: 'type',
+      fixed:true,
```
powerfooI, line 73: Why there's no space between colon and `true` here?

ui/src/pages/Alert/helper.ts
```diff
@@ -144,9 +146,11 @@ export const getInstancesFromRes = (
   if (types.includes('observer')) {
     res.type = 'observer';
     res.observer = getInstanceValues('observer');
+    res.obcluster = [resInstances[0].obcluster!];
```
powerfooI, line 149: Is the length > 0 definitely?

powerfooI (changes requested):

ui/src/pages/Alert/AlarmFilter/index.tsx
yang1666204: I know, but `list` is a union type, and its type is determined by `type`.

ui/src/pages/Alert/helper.ts
yang1666204: Yes.

powerfooI (approved):
LGTM

PR #434, opened 2024-06-07
Add InputTimeComp and fix some bugs

yang1666204:
## Summary
1. Add a custom form component `InputTimeComp` to support selecting time units.
2. Add a drawer component instead of jumping to other pages.
3. Fix some bugs.

### DIFF
--- a/ui/src/pages/Alert/Event/index.tsx
+++ b/ui/src/pages/Alert/Event/index.tsx
@@ -22,13 +22,22 @@ import type { ColumnsType } from 'antd/es/table';
 import dayjs from 'dayjs'
 import AlarmFilter from '../AlarmFilter';
 import { sortEvents } from '../helper';
+import RuleDrawerForm from '../Rules/RuleDrawerForm';
+import { useState } from 'react';
 const { Text } = Typography;
 
 export default function Event() {
   const [form] = Form.useForm();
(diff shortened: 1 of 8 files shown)


ui/src/pages/Alert/Event/index.tsx
```diff
@@ -22,13 +22,22 @@ import type { ColumnsType } from 'antd/es/table';
...
 const { Text } = Typography;
 
 export default function Event() {
   const [form] = Form.useForm();
+  const [drawerOpen,setDrawerOpen] = useState<boolean>(false);
+  const [editRuleName, setEditRuleName] = useState<string>();
   const { data: listAlertsRes, run: getListAlerts } = useRequest(
     alert.listAlerts,
   );
+  const editRule = (rule:string)=>{
```
powerfooI, line 36: This file seems not to be formatted.

powerfooI (changes requested):

ui/src/pages/Alert/Event/index.tsx
yang1666204: Maybe I need a tool that automatically formats before submitting.

powerfooI: Yes, you'd better add a step in husky hook before real commit.

powerfooI (approved):
LGTM

PR #437, opened 2024-06-12
update config in charts, support specify id when upgrade route

chris-sun-star:
## Summary

fix configs for chart of oceanbase-dashboard
specify id when upgrade alarm route

### DIFF
--- a/charts/oceanbase-dashboard/templates/prom-rule-config.yaml
+++ b/charts/oceanbase-dashboard/templates/prom-rule-config.yaml
@@ -8,7 +8,7 @@ metadata:
     name: {{ .Release.Name }}-prometheus-rules-conf
 data:
   prometheus.rules: |-
-    groups:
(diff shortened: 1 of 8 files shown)


charts/oceanbase-dashboard/templates/prom-rule-config.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     name: {{ .Release.Name }}-prometheus-rules-conf
 data:
   prometheus.rules: |-
-    groups:
+    {{`groups:
```
powerfooI, line 11: What is the usage of this pair of `{{` and `}}`

chris-sun-star: to escape all the template marker {{ xx }}  in the config

powerfooI (approved):
LGTM

PR #439, opened 2024-06-12
Rule default

chris-sun-star:
## Summary

set default value of alarm rule's type
clean yum cache files after install mysql in dashboard's dockerfile, it will reduce the image size about 60M


powerfooI (approved):
LGTM

PR #452, opened 2024-06-14
Fix some bugs

yang1666204:
## Summary

1. Fix the problem of wrong request timing.
2. Fix the text display error of `palceholder`.
3. Disable editing of rule names when editing rule configuration.
4. Optimize the display of the list.

### DIFF
--- a/ui/src/pages/Alert/Rules/RuleDrawerForm.tsx
+++ b/ui/src/pages/Alert/Rules/RuleDrawerForm.tsx
@@ -124,53 +124,41 @@ export default function RuleDrawerForm({
           <Col span={16}>
             <Form.Item
               name={'name'}
-              rules={
-                isEdit
-                  ? [
-                      {
-                        required: true,
-                        message: intl.formatMessage({
-                          id: 'src.pages.Alert.Rules.B7B764AE',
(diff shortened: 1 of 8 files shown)


ui/src/pages/Alert/Rules/RuleDrawerForm.tsx
```diff
@@ -124,53 +124,41 @@ export default function RuleDrawerForm({
...
+                  },
+                },
+              ]}
               label={intl.formatMessage({
                 id: 'src.pages.Alert.Rules.14235DA8',
                 defaultMessage: '告警规则名',
               })}
             >
               <Input
+                disabled={isEdit}
```
powerfooI, line 161: Why disable the `Input` when `isEdit` is `true`?

powerfooI (changes requested):

ui/src/pages/Alert/Rules/RuleDrawerForm.tsx
yang1666204: Modifying the rule name will add a new rule. The backend determines the uniqueness of the rule by the rule name.

powerfooI (approved):

PR #458, opened 2024-06-18
fix duplicated silence labels and inactive observer rule config

chris-sun-star:
## Summary

fix duplicated silencer label config
fix alarm rule config for inactive servers


powerfooI (approved):
LGTM

PR #468, opened 2024-06-24
Resole some bugs and optimize the code

yang1666204:
## Summary

1. When selecting `tenants` or `observers`, filter out the corresponding cluster.(ui/src/pages/Alert/helper.ts)
image

3. Add some mock data to reproduce the defect.
4. Remove `Col` from the public component.
5. format code.


ui/src/pages/Alert/Shield/ShieldObjInput.tsx
```diff
@@ -145,28 +149,32 @@ export default function ShieldObjInput({
...
-              maxCount={1}
-              allowClear
-              options={getOptionsFromType(clusterFormName[1])}
-              placeholder={intl.formatMessage({
-                id: 'src.pages.Alert.Shield.B0133BD9',
-                defaultMessage: '请选择集群',
-              })}
-            />
+          <Form.Item noStyle dependencies={[nextFormName]}>
+            {() => (
```
powerfooI, line 153: Why use `{() => <Component />}` instead of `<Component />` here?

ui/src/pages/Cluster/Detail/Overview/ZoneTable.tsx
```diff
@@ -137,8 +137,8 @@ export default function ZoneTable({
   };
   const handleDelete = async (zoneName: string) => {
     const res = await deleteObzoneReportWrap({
-      ns,
-      name,
+      ns:ns!,
```
powerfooI, line 140: Add spaces here.

ui/src/pages/Cluster/New/index.tsx
```diff
@@ -103,11 +103,13 @@ export default function New() {
         style={{ marginBottom: 56 }}
       >
         <Row gutter={[16, 16]}>
-          <BasicInfo
-            passwordVal={passwordVal}
-            setPasswordVal={setPasswordVal}
-            form={form}
-          />
+          <Col span={24}>
```
powerfooI, line 106: Why add `Col` here and delete `Col` somewhere else.

ui/src/pages/Tenant/Detail/Backup/index.tsx
```diff
@@ -1,35 +1,37 @@
...
-            backupPolicyRefresh();
-          },REFRESH_TENANT_TIME)
+  const { refresh: backupPolicyRefresh, loading } = useRequest(
+    getBackupPolicy,
+    {
+      defaultParams: [{ ns, name }],
+      onSuccess: ({ successful, data }) => {
+        if (successful) {
+          setBackupPolicy(data);
+          if (!BACKUP_RESULT_STATUS.includes(data.status)) {
```
powerfooI, line 26: Add comparison `timerRef.current != null` here.

ui/src/pages/Tenant/index.tsx
```diff
@@ -1,73 +1,81 @@
...
+    data: tenantsListResponse,
+    refresh: reGetAllTenants,
+    loading,
+  } = useRequest(getAllTenants, {
+    onSuccess: ({ data, successful }) => {
+      if (successful) {
+        const operatingTenant = data.find(
+          (tenant) => !RESULT_STATUS.includes(tenant.status),
+        );
+        if (operatingTenant) {
```
powerfooI, line 32: Add comparison for `timerRef.current` and `null` here.

powerfooI (changes requested):

ui/src/pages/Alert/Shield/ShieldObjInput.tsx
yang1666204: If you use `<Component />` directly, it will not be rendered. Maybe in `antd`'s view, these two ways of writing are different.
image

ui/src/pages/Cluster/New/index.tsx
yang1666204: Original writing:`<Row><Component /></Row> function Component(){ return <Col>....</Col> }` 
I think `Component` is a public component, and `Col` should be handled by its parent component.

powerfooI (approved):
LGTM

PR #472, opened 2024-06-26
Solve the problem of selecting masked objects

yang1666204:
## Summary

1. Manually add the default tenant sys.
image

3. Add restrictions: Only allow tenants or observers in the current cluster to be selected.
image

5. Get the tenant name from the property `tenantName` instead of `name`.


powerfooI (approved):
LGTM

PR #481, opened 2024-07-04
add priority for observer deletion fix #410

### LINKED ISSUE #410: [Feature]: Delete a specific observer
### Describe your use case

observer which is the combination of a pod running observer process and also the storage holding all the data, sometimes there is some problem with the storage, or the node host the observer. in these situations, add another observer and delete the abnormal one is a common procedure.

### Describe the solution you'd like

annotate the observer to tell ob-operator this observer has a higher priority to be deleted and then modify the replica of the zone this observer belongs to, ob-operator will delete the observer with the highest delete-priority.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

chris-sun-star:
## Summary

add annotation to hint priority for observer deletion
delete observer with the highest deletion priority

### DIFF
--- a/internal/const/oceanbase/misc.go
+++ b/internal/const/oceanbase/misc.go
@@ -13,5 +13,5 @@ See the Mulan PSL v2 for more details.
 package oceanbase
 
 const (
-	DefaultDeletePriority = 100
+	DefaultDeletionPriority = 100
 )
--- a/internal/const/oceanbase/misc.go
+++ b/internal/const/oceanbase/misc.go
@@ -13,8 +13,5 @@ See the Mulan PSL v2 for more details.
 package oceanbase
 
 const (
-	LogLevelDefault = 0
-	LogLevelInfo    = 0
-	LogLevelDebug   = 1
-	LogLevelTrace   = 2
+	DefaultDeletePriority = 100
 )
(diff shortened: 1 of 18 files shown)


internal/resource/obclusteroperation/obclusteroperation_task.go
```diff
@@ -157,6 +158,37 @@ func ModifyClusterSpec(m *OBClusterOperationManager) tasktypes.TaskError {
 				obcluster.Spec.Parameters = append(obcluster.Spec.Parameters, apitypes.Parameter{Name: k, Value: v})
 			}
 		}
+	case constants.ClusterOpTypeDeleteOBServers:
```
powerfooI, line 161: If user pass exact names of observers, why not delete them directly and update replicas of zones?

internal/const/oceanbase/misc.go
```diff
@@ -13,8 +13,5 @@ See the Mulan PSL v2 for more details.
 package oceanbase
 
 const (
-	LogLevelDefault = 0
-	LogLevelInfo    = 0
-	LogLevelDebug   = 1
-	LogLevelTrace   = 2
+	DefaultDeletePriority = 100
```
powerfooI, line 16: Maybe `DeletionPriority` the same as other places is better.

internal/resource/obclusteroperation/obclusteroperation_task.go
chris-sun-star: The problem is these are two resource managed by different controllers, if you'd like to do that way, you may add a annotation to a specific observer and modify the replica number of obzone. Adding this operation is mainly for simplicity and to use it in cli tool or dashboard.

powerfooI (approved):
LGTM

PR #492, opened 2024-07-25
use Ubi as base image, start process as non-root user

chris-sun-star:
## Summary

use Ubi as base image, start process as non-root user

### DIFF
--- a/build/Dockerfile.dashboard
+++ b/build/Dockerfile.dashboard
@@ -14,12 +14,18 @@ COPY . .
 RUN make dashboard-dep-install dashboard-build
 
 # start build docker image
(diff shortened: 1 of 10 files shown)


powerfooI (approved):
LGTM

The size of ubi is more than 400 MB, which introduces a big increase of manifests size.

chris-sun-star:
> LGTM
> 
> The size of ubi is more than 400 MB, which introduces a big increase of manifests size.

yes, but this is necessary to pass openshift's certification.

powerfooI:
Sorry, I had looked a wrong image card. The true size of `ubi` is about 80 MB.

Issue #131, opened 2023-12-01
[Feature]: A powerful cli tool


chris-sun-star:
### Describe your use case

A powerful cli tool to simplify user create and maintain resources

### Describe the solution you'd like

Add a cli tool implementing all the features provided by ob-operator

powerfooI:
We are glad to pronounce the first release of ob-operator CLI tool as `okctl` [here](https://github.com/oceanbase/ob-operator/releases/tag/cli-0.1.0)! 

Great thanks to @lizzy-0323 for continuous devotion for the development of `okctl`.

PR #496, opened 2024-07-29
feat: cli: add cluster management

lizzy-0323:
## Summary

This PR is for #131 

## Solution Description

I have implemented cluster management cmd demo,  but still  not fully completed and tested yet.
For Example:
- Scale Command：No  obcluster operations recorded
- Update command: Not support Storage update
- Show command: Not show obcluster operations about the target obcluster

Build：
```bash
make cli-build 
```


internal/cli/cmd/cluster/update.go
```diff
@@ -0,0 +1,90 @@
...
+			if obcluster.Status.Status != clusterstatus.Running {
+				logger.Fatalln(fmt.Errorf("Obcluster status invalid, Status:%s", obcluster.Status.Status))
+			}
+			if o.Cpu != 0 {
+				obcluster.Spec.OBServerTemplate.Resource.Cpu = *apiresource.NewQuantity(o.Cpu, apiresource.DecimalSI)
+			}
+			if o.MemoryGB != 0 {
+				obcluster.Spec.OBServerTemplate.Resource.Memory = *apiresource.NewQuantity(o.MemoryGB*constant.GB, apiresource.BinarySI)
+			}
+			cluster, err := clients.UpdateOBCluster(cmd.Context(), obcluster)
```
powerfooI, line 63: Should not update obcluster directly here. Created OBClusterOperation will do that instead.

internal/cli/cmd/cluster/upgrade.go
```diff
@@ -0,0 +1,78 @@
...
+				logger.Fatalln(err)
+			}
+			obcluster, err := clients.GetOBCluster(cmd.Context(), o.Namespace, o.Name)
+			if err != nil {
+				logger.Fatalln(err)
+			}
+			if obcluster.Status.Status != clusterstatus.Running {
+				logger.Fatalln(fmt.Errorf("Obcluster status invalid, Status:%s", obcluster.Status.Status))
+			}
+			obcluster.Spec.OBServerTemplate.Image = o.Image
```
powerfooI, line 53: Do not update cluster directly here. The same as command in `update.go`

internal/cli/pkg/cluster/util.go
```diff
@@ -0,0 +1,156 @@
...
+	// 遍历密码中的每个字符
+	for _, char := range password {
+		switch {
+		case regexp.MustCompile(`[A-Z]`).MatchString(string(char)):
+			countUppercase++
+		case regexp.MustCompile(`[a-z]`).MatchString(string(char)):
+			countLowercase++
+		case regexp.MustCompile(`[0-9]`).MatchString(string(char)):
+			countNumber++
+		default:
```
powerfooI, line 73: White space characters will also be counted here. Maybe a determined list of special character instead of `default` is better here.

internal/cli/pkg/cluster/util.go
```diff
@@ -0,0 +1,156 @@
...
+func generateRandomPassword() string {
+	const (
+		maxLength      = 32
+		minUppercase   = 2
+		minLowercase   = 2
+		minNumber      = 2
+		minSpecialChar = 2
+	)
+
+	characters := "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789~!@#%^&*_-+=|(){}[]:;,.?/`$\"<>"
```
powerfooI, line 113: It's suggested to remove some characters that may cause escaping in shell environment like ", `, ' , etc.

internal/cli/pkg/cluster/util.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		if len(sb.String()) >= maxLength {
+			return generateRandomPassword()
+		}
+	}
+
+	return sb.String()
+}
+func generateUUID() string {
+	parts := strings.Split(uuid.New().String(), "-")
```
powerfooI, line 154: You can use `"k8s.io/apimachinery/pkg/util/rand".String()` to generate random alphanumeric string.

internal/clients/obcluster.go
```diff
@@ -89,7 +89,27 @@ func GetOBCluster(ctx context.Context, namespace, name string) (*v1alpha1.OBClus
 func DeleteOBCluster(ctx context.Context, namespace, name string) error {
 	return ClusterClient.Delete(ctx, namespace, name, metav1.DeleteOptions{})
 }
-
+func CreateOBClusterOperation(ctx context.Context, obclusterOperation *v1alpha1.OBClusterOperation) (*v1alpha1.OBClusterOperation, error) {
```
powerfooI, line 92: It seems that there is no empty line in several files containing new functions. Had better add an empty line for separating function blocks in future.

internal/cli/pkg/cluster/create.go
```diff
@@ -0,0 +1,331 @@
+/*
```
powerfooI, line 1: The package `pkg` is redundant which can be removed or renamed to another meaningful one.

powerfooI:
Great job! There are some small pieces that we could discuss about.

make/cli.mk
```diff
@@ -0,0 +1,23 @@
...
+PWD ?= $(shell pwd)
+
+.PHONY: cli-build
+cli-build: cli-dep-install # Build oceanbase-cli
+	go build -o bin/obocli cmd/cli/main.go
+
+.PHONY: cli-clean
+cli-clean: # Clean build
+	rm -rf bin/cli
+	go clean -i ./..
```
powerfooI, line 14: The recursive path should written as `./...` in go's tools.

internal/cli/pkg/cluster/util.go
lizzy-0323: Yes,  I thick maybe use `[^\w\s]` instead?

internal/cli/pkg/cluster/create.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	// if not specific name, using cluster name
+	if o.Name == "" {
+		o.Name = o.ClusterName
+	}
+	return nil
+}
+
+// AddZoneFlags adds the zone-related flags to the command.
+func (o *CreateOptions) AddZoneFlags(cmd *cobra.Command) {
```
chris-sun-star, line 277: maybe it's more reasonable to move functions deal with flags into another package

internal/cli/pkg/cluster/generic.go
```diff
@@ -0,0 +1,19 @@
...
+	http://license.coscl.org.cn/MulanPSL2
+
+THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+See the Mulan PSL v2 for more details.
+*/
+package cluster
+
+type BaseOptions struct {
```
chris-sun-star, line 16: maybe it's better to call it ResourceOptions since it's the unique meta info of a K8s resource

internal/cli/pkg/cluster/scale.go
```diff
@@ -0,0 +1,59 @@
...
+	return nil
+}
+func (o *ScaleOptions) Validate() error {
+	// Ensure obcluster has at least 2 zones
+	deleteNum := 0
+	for _, zone := range o.Topology {
+		if zone.Replicas == 0 {
+			deleteNum++
+		}
+		if o.ZoneNum-deleteNum < 2 {
```
chris-sun-star, line 53: the condition should be the remaining zone number is less than the majority of all zones, 2 is only for zone number 3, if you have 5 zones, this will be 3

chris-sun-star (commented):
Great work, but still need to mention package stuctrues and comments need to be written in English

internal/cli/pkg/cluster/create.go
```diff
@@ -0,0 +1,331 @@
...
+// Parse Cli args and set options
+func (o *CreateOptions) Parse() error {
+	topology, err := mapZonesToTopology(o.Zones)
+	if err != nil {
+		return err
+	}
+	o.Topology = topology
+	// if not specific id, using timestamp
+	if o.ClusterId == 0 {
+		o.ClusterId = time.Now().Unix() % 4294901759
```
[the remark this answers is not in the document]
lizzy-0323: > where does this value 4294901759 come from

After communication，we thick just use random int here.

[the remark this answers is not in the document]
lizzy-0323: > and cluster id can not be 0

This is the default value of this flag, but actually o.ClusterId can not be set as 0.(if 0, then generate random).

internal/cli/cmd/cluster/create.go
```diff
@@ -0,0 +1,57 @@
...
+			obcluster := cluster.CreateOBClusterInstance(o)
+			if err := clients.CreateSecretsForOBCluster(cmd.Context(), obcluster, o.RootPassword); err != nil {
+				logger.Fatalln(errors.Wrap(err, "Create secrets for obcluster"))
+			}
+			_, err := clients.CreateOBCluster(cmd.Context(), obcluster)
+			if err != nil {
+				logger.Fatalln(oberr.NewInternal(err.Error()))
+			}
+			logger.Printf("Create obcluster instance: %s", o.ClusterName)
+			logger.Printf("RootPassword: %s", o.RootPassword)
```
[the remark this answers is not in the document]
lizzy-0323: it will be better if return the k8s command.

internal/cli/pkg/cluster/create.go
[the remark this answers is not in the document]
chris-sun-star: if you calculate the mod of some number, you have chance to get a 0
and I've found this number 4294901759 in OceanBase's document, it's better to define it in a separate file

internal/cli/utils/conf.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"gopkg.in/yaml.v2"
+)
+
+type ComponentVersions struct {
+	Components map[string]string `yaml:"components"`
+}
+
+// filePath for test
+var filePath = "internal/cli/LATEST_VERSION.yaml"
```
powerfooI, line 28: The yaml configuration file could not be bundled into the built binary. Maybe you should refer to `go-bindata` to solve the problem. You can find examples in dashboard `metric_xxx.yaml` in `internel/assets`

internal/cli/cmd/cluster/create.go
```diff
@@ -49,7 +49,7 @@ func NewCreateCmd() *cobra.Command {
 				logger.Fatalln(oberr.NewInternal(err.Error()))
 			}
 			logger.Printf("Create obcluster instance: %s", o.ClusterName)
-			logger.Printf("RootPassword: %s", o.RootPassword)
+			logger.Printf("Run `echo $(kubectl get secret %s -o jsonpath='{.data.password}'|base64 --decode)$` to get the secrets", obcluster.Spec.UserSecrets.Root)
```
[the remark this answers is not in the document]
lizzy-0323: sry, this may be because I typed one more character $

powerfooI (approved):
LGTM

PR #515, opened 2024-08-13
Add support for custom schedulerName in observer pod configuration

qchenzi:
**Description**:

This MR introduces a new feature that allows users to specify a custom scheduler for the observer pods in the ob cluster. The changes include:

- podFields support: A new field podFields has been added to the observer configuration in the CRD, allowing users to define specific configurations related to the pod.
- Custom schedulerName: Under podFields, users can now specify the schedulerName, enabling them to use a custom Kubernetes scheduler instead of the default one.

**Changes include:**

- Updated CRD schema to include podFields and schedulerName.
- Modified the controller logic to apply the schedulerName to the observer pods if specified.
- Ensured backward compatibility so that if schedulerName is not provided, the default scheduler is used.

**Usage Example:**
Here’s an example of how users can define the schedulerName for observer pods using the new podFields field:

```yaml
apiVersion: oceanbase.oceanbase.com/v1alpha1
kind: OBCluster
metadata:
  name: test
  namespace: oceanbase
spec:
  observer:
    image: oceanbase/oceanbase-cloud-native:4.2.3.1-101000032024061316
    podFields:
      schedulerName: custom-scheduler
    resource:
      cpu: 2
      memory: 8Gi
    storage:
      dataStorage:
        ...
      redoLogStorage:
        ..
      logStorage:
        ..
...
```

### DIFF
--- a/internal/resource/utils/podfields.go
+++ b/internal/resource/utils/podfields.go
@@ -1,0 +1,28 @@
+/*
+Copyright (c) 2024 OceanBase
+ob-operator is licensed under Mulan PSL v2.
+You can use this software according to the terms and conditions of the Mulan PSL v2.
+You may obtain a copy of Mulan PSL v2 at:
+         http://license.coscl.org.cn/MulanPSL2
+THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+See the Mulan PSL v2 for more details.
+*/
+
+package utils
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/oceanbase/ob-operator/api/types"
+)
+
+// GetSchedulerName returns the scheduler name from the given PodFields.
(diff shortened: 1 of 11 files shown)


internal/resource/utils/podfields.go
```diff
@@ -0,0 +1,27 @@
...
+THIS SOFTWARE IS PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES OF ANY KIND,
+EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+See the Mulan PSL v2 for more details.
+*/
+
+package utils
+
+import (
+	"github.com/oceanbase/ob-operator/api/types"
```
powerfooI, line 16: The import block seems to have conflicts with `go-imports` lint rule.

powerfooI (commented):
It's quite a great work! The change is so clear!

There is only one lint problem in file `utils/podfields.go`. You can run `make lint` locally to see where it is. :) 

Look forward to your reply 👍🏼

internal/resource/utils/podfields.go
qchenzi: Thank you for the review and the kind words! I appreciate you pointing out the lint issue in utils/podfields.go. I'll run make lint locally to identify and fix the problem as suggested.

powerfooI (approved):
LGTM

PR #516, opened 2024-08-14
access module bugfix

yang1666204:
## Summary
Debug the permission management module interface and fix interaction issues

### DIFF
--- a/ui/config/config.ts
+++ b/ui/config/config.ts
@@ -17,5 +17,5 @@ export default defineConfig({
   },
   routes,
   history: { type: 'hash' },
-  npmClient: 'yarn',
+  npmClient: 'yarn'
 });
--- a/ui/config/config.ts
+++ b/ui/config/config.ts
@@ -17,5 +17,10 @@ export default defineConfig({
   },
   routes,
   history: { type: 'hash' },
-  npmClient: 'yarn'
+  npmClient: 'yarn',
+  proxy:{
+    '/api/v1':{
+      target:"http://11.161.204.4:30681"
+    }
+  }
 });
--- a/ui/config/config.ts
+++ b/ui/config/config.ts
@@ -18,9 +18,4 @@ export default defineConfig({
   routes,
   history: { type: 'hash' },
   npmClient: 'yarn',
-  proxy:{
-    '/api/v1':{
-      target:"http://11.161.204.4:30681"
-    }
-  }
 });
--- a/ui/src/components/customModal/MyInfoModal.tsx
+++ b/ui/src/components/customModal/MyInfoModal.tsx
@@ -1,0 +1,50 @@
+import { useModel } from '@umijs/max';
+import type { DescriptionsProps } from 'antd';
+import { Descriptions } from 'antd';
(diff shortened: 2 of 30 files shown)


ui/config/config.ts
```diff
@@ -17,5 +17,10 @@ export default defineConfig({
   },
   routes,
   history: { type: 'hash' },
-  npmClient: 'yarn'
+  npmClient: 'yarn',
+  proxy:{
+    '/api/v1':{
+      target:"http://11.161.204.4:30681"
```
powerfooI, line 23: The temporary proxy in devServer should not be submitted to the VCS.

ui/src/components/customModal/MyInfoModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+    {
+      key: 'roles',
+      label: '角色',
+      children: `${
+        accountInfo?.roles.map((role) => role.name).join(',') || '-'
+      } `,
+    },
+    {
+      key: 'lastLoginAt',
+      label: '最近一次登陆',
```
powerfooI, line 33: Had better change the label to `最后一次登录时间`. And the `log in` should be translated into `登录` instead of `登陆`.

ui/src/components/customModal/MyInfoModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+      children: `${accountInfo?.lastLoginAt || '-'}`,
+    },
```
powerfooI, line 35: There should be an item that shows description of the account.

ui/src/components/customModal/MyInfoModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+  ];
+  return (
+    <CustomModal
+      isOpen={visible}
+      handleOk={() => {
```
powerfooI, line 40: Can we keep only one button here and set its text content to "OK"?

ui/src/pages/Access/Accounts.tsx
```diff
@@ -82,6 +82,7 @@ export default function Accounts({
       title: '最近一次登陆时间',
```
powerfooI, line 82: Had better modify `登陆` to `登录`

ui/src/pages/Login/index.tsx
```diff
@@ -17,9 +17,12 @@ const Login: React.FC = () => {
   const onFinish = async (values: API.User) => {
     values.password = encryptText(values.password, publicKey) as string;
     const res = await user.login(values);
-    
     if (res.successful) {
-      navigate('/overview');
+      if (res.data.needReset) {
+        navigate('/reset');
+      } else {
+        navigate('/overview');
```
powerfooI, line 24: Can the user replace `xxx/reset` to `xxx/overview` to skip the reset step?

powerfooI (changes requested):
Well done for the most parts! There is a problem we should discuss on that permission enforcer in format of `access.[domain][action]` won't work correctly any more after we add object-granularity control in next several releases.

PR #522, opened 2024-08-22
Enforce read permission if write permission is on

powerfooI:
## Summary

1. Let read request go if the subject has write permission for that object
2. Updated test cases for `rbac` module


chris-sun-star (approved):
LGTM

PR #524, opened 2024-08-22
fix:permission

yang1666204:
## Summary
Fixed the issue where users with "write" permissions did not have "read" permissions.


powerfooI (approved):
LGTM

PR #525, opened 2024-08-26
Some bugfix

yang1666204:
## Summary

1. Fix the issue that role is not updated.
2. Verify that the role exists when creating it.
3. Support administrators to modify other users' passwords.
4. Deleting a role that is already in use by a user is not allowed.

### DIFF
--- a/ui/src/pages/Access/Accounts.tsx
+++ b/ui/src/pages/Access/Accounts.tsx
@@ -27,6 +27,9 @@ export default function Accounts({
   const [editData, setEditData] = useState<AcAccount>();
   const [modalVisible, setModalVisible] = useState<boolean>(false);
   const [resetModalVisible, setResetModalVisible] = useState<boolean>(false);
+  const [showOriginPwd, setShowOriginPwd] = useState(true);
+  const [targetUser, setTargetUser] = useState('');
+
   const { run: deleteAccount } = useRequest(accessReq.deleteAccount, {
     manual: true,
     onSuccess: ({ successful }) => {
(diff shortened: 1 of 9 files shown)


ui/src/pages/Access/index.tsx
```diff
@@ -58,6 +66,16 @@ export default function Access() {
     }
   };
 
+  const existingRole = useMemo(() => {
```
powerfooI, line 69: Naming the variable to plural form like `existingRoles` would be better.

ui/src/pages/Access/Roles.tsx
```diff
@@ -120,6 +130,16 @@ export default function Roles({ allRoles, refreshRoles }: RolesProps) {
     setModalVisible(true);
   };
 
+  const existingRole = useMemo(() => {
```
powerfooI, line 133: Why not pass the `existingRole` from `Access/index` to `Access/Roles`? Actually you've calculated one in parent component.

ui/src/pages/Access/Accounts.tsx
```diff
@@ -122,7 +125,18 @@ export default function Accounts({
         return (
           <Space>
             <Button
-              onClick={() => setResetModalVisible(true)}
+              onClick={() => {
```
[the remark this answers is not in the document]
yang1666204: Okey.

ui/src/pages/Access/index.tsx
yang1666204: Done.

ui/src/pages/Access/Roles.tsx
yang1666204: Done.

powerfooI (approved):
LGTM

PR #527, opened 2024-08-27
fix(action): added latest tag to oceanbase-todo image

powerfooI:


.github/workflows/release-dashboard.yml
```diff
@@ -61,6 +61,9 @@ jobs:
             ${{ vars.DOCKER_PUSH_BASE }}/oceanbase-dashboard:${{ steps.set_version_vars.outputs.version }}
             quay.io/${{ vars.QUAY_IO_PUSH_BASE }}/oceanbase-dashboard:${{ steps.set_version_vars.outputs.version }}
             ghcr.io/${{ github.repository_owner }}/oceanbase-dashboard:${{ steps.set_version_vars.outputs.version }}
+            ${{ vars.DOCKER_PUSH_BASE }}/oceanbase-dashboard:latest
```
chris-sun-star, line 64: Maybe it's not a good idea to use latest image for ob-operator and dashboard, these container require corresponding resources such as CRD or secrets, given the latest image may cause problem if the image updates and the user pulls the new image

chris-sun-star (approved):

PR #528, opened 2024-08-27
Optimize CustomModal

yang1666204:
## Summary

1. Optimize component CustomModal parameter passing.
2. Fix some path issues.

### DIFF
--- a/ui/src/pages/Login/index.tsx
+++ b/ui/src/pages/Login/index.tsx
@@ -18,11 +18,14 @@ const Login: React.FC = () => {
     values.password = encryptText(values.password, publicKey) as string;
     const res = await user.login(values);
     if (res.successful) {
-      if (res.data.needReset) {
-        navigate('/reset');
-      } else {
-        navigate('/overview');
-      }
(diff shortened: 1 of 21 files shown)


ui/src/pages/Login/index.tsx
```diff
@@ -18,11 +18,14 @@ const Login: React.FC = () => {
...
-        navigate('/overview');
-      }
+      // Set a timer to wait for permissions to update
+      setTimeout(() => {
+        if (res.data.needReset) {
+          navigate('/reset');
+        } else {
+          navigate('/overview');
+        }
+      }, 500);
```
powerfooI, line 28: Why do we need this timer to wait for permissions to be updated? What will happen if no such a timer?

yang1666204: If we don't have this timer, access will not be updated after switching users with different permissions.

yang1666204: Maybe I should add l `loading` to the login button.

powerfooI (approved):
LGTM

PR #529, opened 2024-08-27
fix(scripts): modify labels for obproxy set up from the shell

powerfooI:


scripts/setup-obproxy.sh
```diff
@@ -225,6 +263,9 @@ spec:
               name: "sql"
             - containerPort: 2884
               name: "prometheus"
+          envFrom:
+            - configMapRef:
+                name: $CONFIG_MAP
```
chris-sun-star, line 268: what if CONFIG_MAP is empty

powerfooI: The CONFIG_MAP can not be empty. If `--cm` is given and is empty the script will throw an error. If `--cm` and `--env` are all given, it throws an error. If only `--env` is given, the script create a new config map.

chris-sun-star: but what if none of --env or --cm is passed, I suspect the config map is only created when either of these parameters is passed

chris-sun-star (approved):
LGTM

PR #530, opened 2024-08-28
fix:password input&statistics&createdRoles

yang1666204:
## Summary

1. Using `Input.Password` replace `Input`.
2. Duplicate roles cannot be created.
3. Refresh access after editing an account.


powerfooI (approved):
LGTM

PR #542, opened 2024-09-01
feat: cli: add component installation and update

lizzy-0323:
## Summary

refer to #131 

## Solution Description
Add cli installation and update, currently support ob-operator, ob-dashboard, cert-manager, local-path-provisioner

```bash
make cli-build
```


powerfooI (commented):
Is this PR WIP? If so, you'd better mark it as a draft. :)

powerfooI:
It seems that this PR has a lot of commits that belong to previous PR #496. You could rebase this branch onto master branch to simplify the commit history.

internal/assets/cli-templates/component_config.yaml
```diff
@@ -1,6 +1,8 @@
 # LATEST_VERSION
 components:
   cert-manager: "2.2.2_release"
-  # local-path-provisioner: "v0.0.28"
-  # oceanbase-dashboard: "0.2.1"
+  local-path-provisioner: "v0.0.28"
+  local-path-provisioner-dev: "master"
+  ob-dashboard: "0.3.0"
   ob-operator: "2.2.2_release"
+  ob-operator-dev: "master"
```
powerfooI, line 8: `ob-operator-dev` and `local-path-provisioner-dev` should be removed.

lizzy-0323: `dev` means development  version of components, maybe they should be reserved to provide more choice.

lizzy-0323:
> Is this PR WIP? If so, you'd better mark it as a draft. :)

install and update components has been implement.

internal/cli/cmd/install/install.go
```diff
@@ -25,15 +25,32 @@ func NewCmd() *cobra.Command {
...
+- cert-manager: Creates TLS certificates for workloads in Kubernetes and renews the certificates before they expire, ob-operator relies on it for certificate management, which should be installed beforehand.
+
+if not specified, install all the components`,
 		PreRunE: o.Parse,
 		Args:    cobra.MaximumNArgs(1),
 		Run: func(cmd *cobra.Command, args []string) {
-			if err := o.Install(); err != nil {
-				logger.Fatalln(err)
+			if len(args) == 0 {
+				logger.Println("Install all the components")
```
chris-sun-star, line 43: I think it's better to just install the ob-operator and ob-dashboard by default, local-path-provisioner and cert-manager are both 3rd party dependencies, these may already exist on the user's environment.

internal/cli/install/install.go
```diff
@@ -56,30 +55,59 @@ func (o *InstallOptions) Parse(_ *cobra.Command, args []string) error {
...
 			return err
 		}
+		versionFlag := fmt.Sprintf("--version=%s", version)
+		cmd = exec.Command("helm", "install", "oceanbase-dashboard", "ob-operator/oceanbase-dashboard", versionFlag)
+	}
+	return runCmd(cmd)
+}
+
+// preRunCmd preRun two commands for ob-dashboard installation
+func preRunCmd() error {
```
chris-sun-star, line 87: Maybe this function can have a better name, this function is just called when installing dashboard, maybe 'updateHelmRepo' is a better choice

internal/cli/cmd/install/install.go
powerfooI: That's true. `local-path-provisioner` is a weak dependency, and you could check the installation of `cert-manager` by checking existence of resources in `cert-manager.io/v1` like following,

image

lizzy-0323: Ok,  Next, I will only provide installation of ob related components by default.

internal/cli/install/install.go
lizzy-0323: Ok, it has been changed.

internal/cli/install/install.go
```diff
@@ -39,75 +44,136 @@ func (o *InstallOptions) AddFlags(cmd *cobra.Command) {
...
+	cmd := exec.Command("kubectl", "get", "crds", "-o", "name", "|", "grep", "cert-manager")
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		return false
+	}
+
+	// Check if the output contains cert-manager resources
+	expectedResources := []string{
```
powerfooI, line 118: What if a different version of cert-manager has different CRDs?

internal/cli/update/update.go
```diff
@@ -18,16 +18,16 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"github.com/oceanbase/ob-operator/internal/cli/utils"
+	"github.com/oceanbase/ob-operator/internal/cli/install"
 )
 
 type UpdateOptions struct {
```
powerfooI, line 24: How about `type UpdateOptions install.InstallOptions`

internal/cli/install/install.go
lizzy-0323: Fine，maybe I still need  to  run `kubectl api-resources | grep cert-manager.io/v1` to check the environment?

powerfooI: Actually what I mean is that you require all CRDs in the list appear in the output of kubectl command. Would that be too strict?

lizzy-0323: In their official documentation:
`kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.15.3/cert-manager.yaml
`should be run to install all the CRDs.
I think if the user installs this way, maybe all CRDs should be found in the environment?

powerfooI: If the version is fixed to v1.15.3, the requirement is proper. How about other version like v1.14, v1.13 or upcoming v1.16? The list of CRDs should be check if they are the same between versions.

lizzy-0323: I have tried official repo with version of v1.14, v1.13, v1.12 and v1.11, the list of CRDs are the same.

powerfooI: That's great!

powerfooI (approved):
LGTM

PR #543, opened 2024-09-02
Fixed some problems found in testing for dashboard 0.3.1

powerfooI:
1. Forbid to create or patch a role w/o permissions
2. Added ceiling for IOPS of resource unit


chris-sun-star (approved):
See whether we can change the CRD of obtenant later

PR #544, opened 2024-09-03
Adjusted permission required by several api

powerfooI:
1. Fixed wrong configuration of api that creates connection to a tenant.
2. Modify permission required by metrics api from `system:read` to `obcluster:read`.


chris-sun-star (approved):
LGTM

PR #547, opened 2024-09-03
Fine-grained control of some permissions

yang1666204:
## Summary

1. Fine-grained control of some permissions.
2. Iops limit.
3. Optimize the UI of Resource pool Modal.
image
4. Improve the verification of role permissions.
image

### DIFF
--- a/ui/src/components/SelectNSFromItem/index.tsx
+++ b/ui/src/components/SelectNSFromItem/index.tsx
@@ -7,6 +7,7 @@ import type { FormInstance } from 'antd/lib/form';
 import AddNSModal from '@/components/customModal/AddNSModal';
 import { resourceNameRule } from '@/constants/rules';
 import { getNameSpaces } from '@/services';
+import { useAccess } from '@umijs/max';
 import { useState } from 'react';
 export default function SelectNSFromItem({
   form,
@@ -15,8 +16,11 @@ export default function SelectNSFromItem({
 }) {
   // control the modal for adding a new namespace
   const [visible, setVisible] = useState(false);
-  const { data, run: getNS } = useRequest(getNameSpaces);
-
+  const { data, run: getNS } = useRequest(getNameSpaces, {
+    manual: true,
+  });
+  const access = useAccess();
+  const hasSysAccess = access.systemread || access.systemwrite;
   const filterOption = (
     input: string,
     option: { label: string; value: string },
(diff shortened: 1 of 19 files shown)


ui/src/components/SelectNSFromItem/index.tsx
```diff
@@ -25,28 +29,36 @@ export default function SelectNSFromItem({
...
-        >
-          <PlusOutlined />
-          <span style={{ marginLeft: '6px' }}>
-            {intl.formatMessage({
-              id: 'OBDashboard.Cluster.New.BasicInfo.AddNamespace',
-              defaultMessage: '新增命名空间',
-            })}
-          </span>
-        </div>
+        {hasSysAccess ? (
```
powerfooI, line 32: `system:write` is required to create a namespace. The above `access.systemread || access systemwrite` is incorrect.

ui/src/components/customModal/ModifyUnitDetailModal.tsx
```diff
@@ -501,30 +495,30 @@ export default function ModifyUnitDetailModal({
...
+                  defaultMessage: '请输入',
+                })}
+              />
+            </Form.Item>
+          </Col>
+          <Col span={8}>
+            <Form.Item label="max iops" name={['unitConfig', 'maxIops']}>
+              <InputNumber
+                style={{ width: 222 }}
+                min={minResource.maxIops}
```
powerfooI, line 514: Does this mean that `maxIops` can not be decreased?

ui/src/pages/Alert/Shield/index.tsx
```diff
@@ -289,7 +293,7 @@ export default function Shield() {
           </h2>
         }
         extra={
-          access.alarmwrite ? (
+          access.alarmwrite || !access.obclusterread ? (
```
powerfooI, line 296: Why can a user create a shield without `obcluster:read` permission? I thought `access.alarmwrite && access.obclusterread` should be correct.

ui/src/pages/OBProxy/index.tsx
```diff
@@ -53,17 +54,21 @@ export default function OBProxy() {
...
+        ) : null}
       </Row>
-      <MonitorComp
-        filterLabel={[]}
-        queryScope="OBPROXY"
-        type="OVERVIEW"
-        groupLabels={['cluster']}
-        queryRange={DEFAULT_QUERY_RANGE}
-      />
+      {access.systemread || access.systemwrite ? (
```
powerfooI, line 63: Why add `system` permission check here while remove the check in `pages/Cluster/index.tsx`?

ui/src/components/customModal/ModifyUnitDetailModal.tsx
yang1666204: I remember there is a minimum value for resources (minResource). If it is less than this value, the resource pool cannot be created. So the input box here limits the minimum value.

powerfooI: OK

ui/src/pages/OBProxy/index.tsx
yang1666204: This is missing, the check for system should be removed.

powerfooI (approved):
LGTM

PR #550, opened 2024-09-04
Join enforcers with OR and AND logic

powerfooI:
A concrete scenario is users with `obcluster:read` or `obproxy:read` permission should both be able to fetch metrics data.

### DIFF
--- a/internal/dashboard/business/ac/enforcers.go
+++ b/internal/dashboard/business/ac/enforcers.go
@@ -58,3 +58,26 @@ func PathGuard(domain, resource, action string) EnforceFunc {
 		return true, nil
 	}
 }
+
+func unionHelper(allPass bool, enforces ...EnforceFunc) EnforceFunc {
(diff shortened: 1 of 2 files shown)


internal/dashboard/business/ac/enforcers.go
```diff
@@ -58,3 +58,26 @@ func PathGuard(domain, resource, action string) EnforceFunc {
 		return true, nil
 	}
 }
+
+func unionHelper(allPass bool, enforces ...EnforceFunc) EnforceFunc {
```
chris-sun-star, line 62: I think it can be more straight forward to use an enum value indicating 'and' or 'or'

powerfooI: OK, but the `allPass` var is still required for simplicity.

chris-sun-star (approved):
LGTM

PR #588, opened 2024-10-15
Fixed unexpected selection of PVCs

powerfooI:
## Summary

1. Fixed wrong `ref-uid` field and supplemented `ref-observer` field in labels of observers' sub resources like pvc, pod and svc.


chris-sun-star (approved):
LGTM

PR #590, opened 2024-10-16
Fix: bugs in cluster and tenant creating

lizzy-0323:
## Summary

Fix bugs in creating tenant and cluster:
1. cluster: fix wrong default values in flags, add backupVolume support when creating clusters
2. tenant: fix parse function 
3. util: remove some characters in password that maybe escaped

## Solution Description

99f4ba370479c347ae432e2d273cb8668ab74067^..b048580caaab0fa4aaa222cb973de252f1d316ee


powerfooI (approved):
LGTM

PR #598, opened 2024-10-28
 improve Status Chinese display and fix scheduling time display error

TianWuwt:
## Summary

1. improve Status Chinese display
Pasted Graphic 1
Pasted Graphic
Pasted Graphic 2
Pasted Graphic 3
Pasted Graphic 4
Pasted Graphic 5
Pasted Graphic 6
2. fix scheduling time display error


ui/src/constants/index.ts
```diff
@@ -379,3 +367,109 @@ export {
   ZONE_IMG_MAP,
   getMinResource,
 };
+
+export const STATUS_LIST = [
```
powerfooI, line 371: Why no use map directly instead of list if all the usages are paired with `findByValue`?

TianWuwt: The data processed by the map cannot be used directly and is prone to errors.

powerfooI: I think `{"running": {...}}` can be used directly. What if `findByValue` find nothing if you use list? When talking about error-proneness lists are almost the same as maps.

TianWuwt: Different methods, different habits, the same results （＾∀＾）

powerfooI (approved):
LGTM

PR #599, opened 2024-10-29
feat(cli): demo command for easier cluster creation

lizzy-0323:
## Summary

This PR is for #131 
Add demo command for easier cluster creation:
currently support:
- single-node cluster
- three-node(1-1-1) cluster

## Solution Description

Clusters can be deployed interactively as follows:

https://github.com/user-attachments/assets/3415e517-f9d5-44c5-8924-8ece48f50f9a


powerfooI (commented):
LGTM. I suggest that flags used to wait for cluster to be ready and that used to create tenant are also needed.

lizzy-0323:
> I don't think demo needs the two parameters backup address and path. Maybe we should remove them.

Ok. In order to provide more convenient usage scenarios, these parameters will no longer be provided.

lizzy-0323:
When using the demo command to create clusters, add the corresponding tenant, and the tenant will wait synchronously for the cluster to be created. During the creation process, both resources will print the current tasks and their statuses. After creation, log the command for the user to retrieve the password for the corresponding resource, shown as follows: 
https://github.com/user-attachments/assets/66b7a1cc-0070-4002-b9ec-80c83fc39222

chris-sun-star:
> When using the demo command to create clusters, add the corresponding tenant, and the tenant will wait synchronously for the cluster to be created. During the creation process, both resources will print the current tasks and their status, shown as follows:
> 279_1730729699.mp4
What's the tenant's password, is it the same as the cluster's root@sys user, maybe it's better to set the tenant's root password as well.
Is it possible to use the tenant's name as resource name

lizzy-0323:
> > When using the demo command to create clusters, add the corresponding tenant, and the tenant will wait synchronously for the cluster to be created. During the creation process, both resources will print the current tasks and their status, shown as follows:
> > 279_1730729699.mp4
> > What's the tenant's password, is it the same as the cluster's root@sys user, maybe it's better to set the tenant's root password as well.
> > Is it possible to use the tenant's name as resource name

For the first question: yes, after creation, the command for user to get the root password will be logged.
For the second question: I thick maybe not, because the check of the resource name and tenant name seems not consistent🤔, you can refer to `/internal/cli/utils/validator.go` for details.

powerfooI (commented):
LGTM

powerfooI (approved):

PR #603, opened 2024-11-04
fix：After deleting a cluster, clicking the browser back page jumps 

TianWuwt:
… an error

## Summary

After deleting a cluster, clicking the browser back page jumps to an error

## Solution Description

image


powerfooI (approved):
LGTM

PR #605, opened 2024-11-04
Set environment variable for obagent to overwrite host ip config

chris-sun-star:
## Summary

Set environment variable for obagent to overwrite host ip config
1. set HOST_IP to 127.0.0.1 when cluster mode is standalone
2. set HOST_IP to service address when cluster mode is service


powerfooI (approved):
LGTM

PR #607, opened 2024-11-04
build: Add a replace block to the go.mod file

RGaius:
Added the replace block in go.mod to solve the problem that the k8s module could not be loaded during goland IDE development


powerfooI (approved):
LGTM

PR #629, opened 2024-11-13
support tenant parameter management

chris-sun-star:
## Summary

support tenant parameter management
tenant variable management will be in another PR


internal/resource/obtenant/utils.go
```diff
@@ -920,3 +922,91 @@ func CreateUserWithCredentials(m *OBTenantManager) error {
 	}
 	return nil
 }
+
+func (m *OBTenantManager) listOBParameters() (*v1alpha1.OBParameterList, error) {
+	obparameterList := &v1alpha1.OBParameterList{}
+	err := m.Client.List(m.Ctx, obparameterList, client.MatchingLabels{
+		oceanbaseconst.LabelRefOBCluster: m.OBTenant.Spec.ClusterName,
```
powerfooI, line 929: Why not list the parameters with UID field of OBTenant as `RefUID` label?

internal/resource/obtenant/utils.go
```diff
@@ -920,3 +922,91 @@ func CreateUserWithCredentials(m *OBTenantManager) error {
...
+	err := m.Client.Create(m.Ctx, obparameter)
+	if err != nil {
+		m.Logger.Error(err, "create obparameter failed")
+		return errors.Wrap(err, "create obparameter")
+	}
+	return nil
+}
+
+func (m *OBTenantManager) updateOBParameter(parameter *apitypes.Parameter) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
powerfooI, line 979: `RetryOnConflict` will take effect only when the function get an `IsConflict` error which is defined in client-go packages. Maybe we should use more general `OnError` here.

internal/resource/obtenant/utils.go
chris-sun-star: These labels both work, I just follow the logic of obcluster manager.

chris-sun-star: I think it's a good idea to use another label, and do not set the cluster name label when it's a tenant parameter

internal/resource/obtenant/utils.go
chris-sun-star: I think using RetryOnConflict is the right choice, this function is called by tenant manager and update the resource parameter managed by paramenter manager, it's possible to encounter a conflict scenario, if there's other error, we can leave that in another reconcile loop.

internal/resource/obtenant/utils.go
powerfooI: > These labels both work, I just follow the logic of obcluster manager.

Listing parameters with only cluster label will get more data than tenant manager need. I recommend to use tenant `uid` as the filtering label or both of they two. :)

chris-sun-star: Yes, that's right, I have changed the label of obparameter which is created with obtenant, use the label ref-obtenant instead of ref-obcluster, thus when the obcluster manager list the parameters, it will only get the relevant parameters, but I still keep the filter logic.

internal/resource/obtenant/utils.go
powerfooI: Oh sorry, it's right. I thought there are direct SQL statement execution in the function.

powerfooI (approved):
LGTM

PR #632, opened 2024-11-14
Tenant variables

chris-sun-star:
## Summary

1. add api type obtenantvariable and corresponding controller
2. add  variables in obtenant
3. modify make command, install kubebuilder if necessary, add generate before docker-build


config/samples/oceanbase_v1alpha1_obtenantvariable.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: oceanbase.oceanbase.com/v1alpha1
```
powerfooI, line 1: Maybe this file should be removed, which will be unused in our scenario.

internal/const/status/obtenantvariable/obtenantvariable.go
```diff
@@ -0,0 +1,20 @@
...
+EITHER EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO NON-INFRINGEMENT,
+MERCHANTABILITY OR FIT FOR A PARTICULAR PURPOSE.
+See the Mulan PSL v2 for more details.
+*/
+
+package obtenantvariable
+
+const (
+	New       = "new"
+	PendingOB = "pending ob"
```
powerfooI, line 17: What is the meaning of `pending ob`? It's somewhat confusing.

internal/resource/obtenant/obtenant_task.go
```diff
@@ -608,3 +608,39 @@ func MaintainTenantParameters(m *OBTenantManager) tasktypes.TaskError {
...
+		m.Logger.V(oceanbaseconst.LogLevelDebug).Info("Build variable map", "variable", variable.Name)
+		variableMap[variable.Name] = variable
+	}
+	for _, variable := range m.OBTenant.Spec.Variables {
+		variableStatus, variableExists := variableMap[variable.Name]
+		if !variableExists {
+			m.Logger.V(oceanbaseconst.LogLevelDebug).Info("Variable not exists, need create", "variable", variable.Name)
+			err := m.createOBTenantVariable(&variable)
+			if err != nil {
+				m.Logger.Error(err, "Create obtenantvariable failed", "param", variable.Name)
```
powerfooI, line 624: Maybe the `"param"` in the logging message should be replaced with `varaiable` or `var`

internal/resource/obtenant/obtenant_task.go
```diff
@@ -608,3 +608,39 @@ func MaintainTenantParameters(m *OBTenantManager) tasktypes.TaskError {
...
+			}
+		} else if variableStatus.Value != variable.Value {
+			m.Logger.V(oceanbaseconst.LogLevelDebug).Info("Variable value not matched, need update", "variable", variable.Name)
+			err := m.updateOBTenantVariable(&variable)
+			if err != nil {
+				m.Logger.Error(err, "Update obtenantvariable failed", "param", variable.Name)
```
powerfooI, line 630: The same as the above situation.

internal/resource/obtenant/utils.go
```diff
@@ -1010,3 +1013,74 @@ func (m *OBTenantManager) deleteOBParameter(parameter *apitypes.Parameter) error
...
+func (m *OBTenantManager) deleteOBTenantVariable(variable *apitypes.Variable) error {
+	obtenantvariable := &v1alpha1.OBTenantVariable{}
+	err := m.Client.Get(m.Ctx, types.NamespacedName{
+		Namespace: m.OBTenant.Namespace,
+		Name:      m.generateVariableName(variable.Name),
+	}, obtenantvariable)
+	if err != nil {
+		return errors.Wrap(err, "Get obtenantvariable")
+	}
+	obtenantvariable.Spec.Variable.Value = variable.Value
```
powerfooI, line 1080: Setting the value of variable before deleting it seems to be redundant~

pkg/oceanbase-sdk/operation/tenant.go
```diff
@@ -130,15 +130,6 @@ func (m *OceanbaseOperationManager) GetCharset(ctx context.Context) (*model.Char
 	return charset, nil
 }
 
-func (m *OceanbaseOperationManager) GetVariable(ctx context.Context, name string) (*model.Variable, error) {
```
powerfooI, line 133: Is this function unused in previous code?

pkg/helper/converter/converter.go
```diff
@@ -40,3 +41,20 @@ func ConvertFloat(value any) any {
 		return v
 	}
 }
+
+func AutoConvert(value string) any {
```
powerfooI, line 45: What are the usages of this function?

chris-sun-star: When setting the variable values, you need to use the corresponding type, or OceanBase will raise an error, this function try to convert the value in string type which is defined in our CRD to the right type.

internal/const/status/obtenantvariable/obtenantvariable.go
chris-sun-star: This is initially used in obparameter status, pending ob means oceanbase cluster is under operation, this may happen when oceanbase cluster is upgrading, and the upgrade script will modify some parameters and the parameters are supposed not to be modified during the procedure, so we add this status for such scenarios.

chris-sun-star: for variables, this status is not needed, I'll delete it

powerfooI (approved):
LGTM

PR #638, opened 2024-11-18
fix: check if resource exists in `update` cmd

lizzy-0323:
## Summary

Check if the resource exists in the environment, if not, do not install it.


internal/cli/utils/checker.go
```diff
@@ -53,3 +90,38 @@ func CheckTenantRole(tenant *v1alpha1.OBTenant, role apitypes.TenantRole) error
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+
+	if err != nil {
+		return false
+	}
+
+	for _, resource := range resourceList {
+		if !bytes.Contains(out.Bytes(), []byte(resource)) {
```
powerfooI, line 122: Had better call `Bytes()` once outside the loop.

powerfooI (approved):
LGTM

PR #647, opened 2024-11-21
feat(cli): add completion cmd and remove unused flag

lizzy-0323:
## Summary

1. Add completion cmd and bind flags as required.
2. Remove `name` flag in backup related cmds, the reason is that this flag is in need: 3ae4efda8d0c4752aa83a3ba869978e7a6a55bd9

## Solution Description

By using completion cmd to generate config files  in shells, user can use auto-completion to manage resources easily.


powerfooI (approved):
LGTM

PR #660, opened 2024-12-02
Support kube-ovn as cni to provide static ip address

chris-sun-star:
## Summary

Support kube-ovn as cni to provide static ip address


powerfooI (approved):
LGTM

PR #664, opened 2024-12-12
Fix(cli): bugs in cli first round testing

lizzy-0323:
## Summary

This PR fix bugs in the first round of cli testing. The problem solved is as follows:
1. fix version flag panic : 2a1d3633d7540e2cfe55a75ec5bc0bcf529d5f71
2. fix password checking in demo cmd: db048b940b33319d8af74c597aa4cd25f2b92733
3. fix password logging when using specified namespace: 5f03455c0aa4f1b5903a84420fae4f6612270834
4. Add checking whether the tenant exists before establishing the cluster: db048b940b33319d8af74c597aa4cd25f2b92733
5. fix helm error when using helm<3.7.0: 4db89633873452b0a35ed43f93eeb79f8ccb0b3f
6. fix duplicate item in backup-policy 's `show` cmd: 426a61f0e946f2230d381de246d9d3e18f5a7378
7. fix: remove `Complete` function when creating a backup policy: 81a827d196ac9c6caf2aba7e708870dfcd2fa815, and the details can be seen in the commit message. The tenant’s create command has also been modified in 4faa655e7f9670360755d225c00d7307abd6038e
8. fix lint: 13f4779c55234a987f7d7655584b6462a228c3ec


internal/cli/cmd/demo/demo.go
```diff
@@ -40,20 +40,27 @@ func NewCmd() *cobra.Command {
...
+					logger.Fatalln(err)
+				}
+				prompt = pf.CreatePrompt(cluster.FLAG_NAMESPACE)
+				if clusterOptions.Namespace, err = pf.RunPromptE(prompt); err != nil {
+					logger.Fatalln(err)
+				}
+				if !utils.CheckIfClusterExists(cmd.Context(), clusterOptions.Name, clusterOptions.Namespace) {
+					break
+				}
+				logger.Printf("Cluster %s already exists in namespace %s, please input another cluster name", clusterOptions.Name, clusterOptions.Namespace)
```
powerfooI, line 63: This `for-loop` seems to be endless if the flow gets here.

internal/cli/cmd/demo/demo.go
```diff
@@ -63,9 +70,17 @@ func NewCmd() *cobra.Command {
...
+			// prompt for tenant create options
+			for {
+				prompt = pf.CreatePrompt(tenant.FLAG_TENANT_NAME_IN_K8S)
+				if tenantOptions.Name, err = pf.RunPromptE(prompt); err != nil {
+					logger.Fatalln(err)
+				}
+				if !utils.CheckIfTenantExists(cmd.Context(), tenantOptions.Name, clusterOptions.Namespace) {
+					break
+				}
+				logger.Printf("Tenant %s already exists in namespace %s, please input another tenant name", tenantOptions.Name, clusterOptions.Namespace)
```
powerfooI, line 83: The same as the above

internal/cli/cmd/demo/demo.go
lizzy-0323: Yes, because here we want to check if the cluster name typed by user is available in the corresponding namespace, and if it is not exists (which is meet the requirements, the return value of `utils.CheckIfClusterExists` is `false` ), the workflow will be advanced.

powerfooI: OK, I see. The cli tool is asking the user to enter the name and namespace of the cluster here.

powerfooI (approved):

PR #665, opened 2024-12-13
fix: update shorthand for namespace and move shorthands to `enter.go`

lizzy-0323:
## Summary

1. update shorthand for namespace 
2. move shorthands for each command to `enter.go`


powerfooI (approved):
Good Job! LGTM

PR #666, opened 2024-12-14
refactor(cli): install cmd and update cmd

lizzy-0323:
## Summary

Refactor install cmd and update cmd, and solve some bugs found in first round testing at the same time:
  1. When using `update` cmd, component with dev version no longer support
  2. Distinguish between `helm upgrade` and `helm install` commands when using install and update for ob-dashboard

## Solution Description

1. Decouple the code of the install and update commands.
2. place the function of running the command in `cmd_runner.go`.
3. move the function of reading config to `config.go`.


powerfooI (approved):
LGTM

PR #680, opened 2024-12-19
refactor: change the way to obtain component version

lizzy-0323:
## Summary

Now the cli tool don't need a yaml file to obtain component version.

### DIFF
--- a/internal/cli/config/config.go
+++ b/internal/cli/config/config.go
@@ -14,53 +14,117 @@ See the Mulan PSL v2 for more details.
 package config
 
 import (
-	"bytes"
-	"fmt"
+	"errors"
+	"io"
+	"net/http"
 
-	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 
-	"github.com/oceanbase/ob-operator/internal/cli/generated/bindata"
 	"github.com/oceanbase/ob-operator/internal/cli/utils"
 )
 
-// component config for test
-var confPath = "internal/assets/cli-templates/component_config.yaml"
+// Chart struct to parse the chart.yaml
+type Chart struct {
+	apiVersion  string `yaml:"apiVersion"`
+	AppVersion  string `yaml:"appVersion"`
+	description string `yaml:"description"`
+	name        string `yaml:"name"`
+	chartType   string `yaml:"type"`
+	version     string `yaml:"version"`
+}
 
-func readComponentConf(path string) map[string]string {
-	components := make(map[string]string)
-	fileobj, err := bindata.Asset(path)
-	// panic if file not exists
+const (
+	stableVersion = "stable"
+	devVersion    = "master"
+)
+
+// ComponentList is the list of components that can be installed
+var ComponentList = []string{
+	"cert-manager", "ob-operator", "ob-dashboard", "local-path-provisioner", "local-path-provisioner-dev", "ob-operator-dev",
+}
+
+// ComponentUpdateList is the list of components that can be updated
+var ComponentUpdateList = []string{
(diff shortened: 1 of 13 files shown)


internal/cli/config/config.go
```diff
@@ -14,53 +14,117 @@ See the Mulan PSL v2 for more details.
...
+}
+
+// ComponentUpdateList is the list of components that can be updated
+var ComponentUpdateList = []string{
+	"cert-manager", "ob-operator", "ob-dashboard", "local-path-provisioner",
+}
+
+var versionURLs = map[string]string{
+	"ob-dashboard":           "https://raw.githubusercontent.com/oceanbase/ob-operator/refs/heads/stable/charts/oceanbase-dashboard/Chart.yaml",
+	"local-path-provisioner": "https://raw.githubusercontent.com/rancher/local-path-provisioner/refs/tags/v0.0.30/deploy/chart/local-path-provisioner/Chart.yaml",
```
powerfooI, line 53: Maybe we should use `master` branch here.

internal/cli/config/config.go
```diff
@@ -14,53 +14,117 @@ See the Mulan PSL v2 for more details.
...
+	return chart.AppVersion, nil
+}
+
+func getVersion(component string) (string, error) {
+	switch component {
+	case "ob-dashboard", "local-path-provisioner":
+		return getVersionFromChart(component)
+	case "cert-manager", "ob-operator":
+		return stableVersion, nil
+	case "ob-operator-dev", "local-path-provisioner-dev":
```
powerfooI, line 93: I think `local-path-provisioner` has no `dev` version. The chart put in master branch of local-path-provisioner is always for production.

lizzy-0323: Through the readme of `local-path-provisioner` repo it can be seen that they provide `dev` and `stable` version. Do we need to provide only the version under the master branch here?

powerfooI: As far as I see, the chart in the master branch is always of the stable version. The `stable` and `dev` are used to describe the status of `deploy/local-path-storage.yaml`, I thought.

internal/cli/config/config.go
powerfooI: As far as I see, the chart in the master branch is always of the stable version. The stable and dev are used to describe the status of deploy/local-path-storage.yaml, I thought.

internal/cli/config/config.go
powerfooI: If users want to install `local-path-provisioner-dev`, the okctl will install it through `deploy/local-path-storage.yaml` right?

lizzy-0323: Yes, for okctl: 
- `stable`->`https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.30/deploy/local-path-storage.yaml`
- `dev`-> `https://raw.githubusercontent.com/rancher/local-path-provisioner/master/deploy/local-path-storage.yaml`
And for now, `getVersion` function is used to get v0.0.30 for stable version.
Actually, I just found that we can just set stable version to `v0.0.30` temporarily, no need to visit the url and parse the version.

powerfooI: It's not suggested to hardcode the version in code.

powerfooI: Removing local-path-provisioner-dev is a more reasonable choice. Because we are not responsible for what it would happen if users install local-path-provisioner-dev

lizzy-0323:
After discussion, we decide to remove `dev` version of `local-path-provisioner` temporarily.

powerfooI (approved):
LGTM

PR #683, opened 2024-12-19
Add the i18n for UI texts

TianWuwt:
## Summary

Increase the internationalization of copywriting


powerfooI:
See https://github.com/oceanbase/ob-operator/pull/687

PR #704, opened 2025-01-08
check there is indeed an upgrade route

chris-sun-star:


pkg/helper/upgrade.go
```diff
@@ -221,5 +221,13 @@ func FindShortestUpgradePath(nodeMap map[string]*VersionDep, startVersionFull, t
 	if len(res) == 1 {
 		res = append([]VersionDep{*startNode}, res...)
 	}
+	result := make([]string, 0)
+	for _, v := range res {
+		result = append(result, v.Version)
+	}
+	log.Println("final result", result)
+	if len(result) == 0 || result[len(result)-1] != targetVersion {
```
powerfooI, line 229: Will the target version be located at the back of the list?

powerfooI (approved):

PR #722, opened 2025-01-10
Fix get cluster resource

chris-sun-star:
## Summary

dashboard may connect to an observer in another obcluster when get cluster resource


powerfooI (approved):
LGTM

PR #740, opened 2025-01-24
Add chart for obproxy-ce

chris-sun-star:
## Summary

Add chart for obproxy-ce

## Solution Description

1. variable obcluster is required, the chart will extract the info of RS_LIST and PROXYRO_PASSWORD from the obcluster resource. The simplest command to deploy obproxy-ce for obcluster test is as follows.
`helm install test ./obproxy-ce --namespace oceanbase --set obcluster=test`
2. obproxy parameters can be set using `--set parameters.key=value`, multiple parameters can be set by multiple --set parameters like `--set parameters.key1=value1 --set parameters.key2=value2`, for example, if some application requires mysql version >= 8.0, then you can set it using the following command.
`helm install test ./obproxy-ce --namespace oceanbase --set obcluster=test --set parameters.mysql_version=8.0.30`


alv00 (approved):
LGTM

PR #749, opened 2025-02-27
Support set proxyro password when create obcluster

chris-sun-star:
## Summary

Support set proxyro password when create obcluster


LINxiansheng (approved):
LGTM

PR #761, opened 2025-03-12
return version when get obcluster info

chris-sun-star:
## Summary

return version when get obcluster info


LINxiansheng (approved):
LGTM

LINxiansheng (commented):
LGTM

PR #763, opened 2025-03-13
add update and patch operation for basic K8s resources

chris-sun-star:
## Summary

add update and patch operation for basic K8s resources


LINxiansheng (approved):
LGTM

PR #776, opened 2025-03-27
Fix test problems in multi-cluster scenario

chris-sun-star:
## Summary

Fix test problems in multi-cluster scenario


LINxiansheng (approved):
LGTM

PR #778, opened 2025-03-31
Fix test problems

chris-sun-star:
## Summary

1. return meta info of obcluster parameters
2. fix encryption problem when passing kubeConfig
3. add route for deleting remote k8s cluster
4. query remote node metrics


LINxiansheng (approved):
LGTM

PR #782, opened 2025-04-07
ignore connection error when get obcluster detail

chris-sun-star:
## Summary

ignore connections when getting obcluster version


LINxiansheng (approved):
LGTM

PR #795, opened 2025-05-14
fix：修复遥测数据报错展示

TianWuwt:
## Summary

修复遥测接口漏出展示


ui/src/app.tsx
```diff
@@ -26,14 +26,18 @@ export const request: RequestConfig = {
     },
     errorHandler: (err) => {
       console.log('errorHandler', err);
+      const { options: { HIDE_ERROR_MESSAGE } = {} } = err.request || {};
```
chris-sun-star, line 29: the options should be retrieved from err.config

PR #801, opened 2025-05-29
Add http api for monitor service discovery 

chris-sun-star:
## Summary

1. Add http api for monitor service discovery, add an api return all the endpoints.
2. Monitor the prometheus config file to use http for service discovery.


LINxiansheng (approved):
LGTM

PR #815, opened 2025-07-18
Support obcluster inspection

chris-sun-star:


chris-sun-star:
The lint problem seems like it's a mistake, the code compiles and works normally, yaml is imported using "gopkg.in/yam.v2"

chris-sun-star:
The lint problem seems like it's a mistake, the code compiles and works normally, yaml is imported using "gopkg.in/yam.v2"

cheer-chen (approved):

PR #827, opened 2025-07-29
set owner references for triggered cronjob

chris-sun-star:
## Summary

set owner references for triggered cronjob


cheer-chen (approved):
LGTM

PR #828, opened 2025-07-29
return successful inspection result as well

chris-sun-star:
## Summary

return successful inspection result as well


cheer-chen (approved):
LGTM

PR #929, opened 2026-04-23
feat: integrate obproxy

lightzt99:


chris-sun-star:
I don't think it's necessary to add a CRD for obproxy, if you'd like to get the rs_list info from kubernetes, you can simply add it into obproxy's image, and a better approach would be to implement the rs_list fetching logic from K8s in this project https://github.com/oceanbase/ob-configserver, it's always a better practice to use ob-configserver rather than using rs_list for obproxy.

PR #933, opened 2026-06-12
feat: support to deploy oblogservice and oceanbase.ai

Junkrat77:
## Summary

Support deploying oceanbase.ai with OBLogService managed by ob-operator.

## Solution Description

- Add `OBLogServiceCluster`, `OBLogServiceZone`, and `OBLogServiceNode` CRDs and controllers.
- Add LogService bootstrap, scaling, recovery, and node lifecycle management.
- Add `shared_storage` deployment mode, shared-storage configuration, and LogService reference to `OBCluster`.
- Integrate shared-storage setup into OceanBase bootstrap and zone management.
- Update the CLI, Dashboard, webhooks, RBAC, Helm chart, and deployment manifests.


api/v1alpha1/obcluster_webhook.go
```diff
@@ -183,7 +187,22 @@ func (r *OBCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, erro
...
 	if existOld && exist && oldMode != mode {
 		return nil, errors.New("mode cannot be changed")
-	} else if !oldCluster.SupportStaticIP() && (oldResource.Cpu != newResource.Cpu || oldResource.Memory != newResource.Memory) {
+	}
+	if oldCluster.Spec.DeploymentMode != r.Spec.DeploymentMode {
+		return nil, errors.New("deploymentMode cannot be changed after creation")
+	}
+	if r.Spec.DeploymentMode == oceanbaseconst.DeploymentModeSharedStorage {
+		if oldCluster.Spec.OBServerTemplate.Image != r.Spec.OBServerTemplate.Image {
+			return nil, errors.New("shared_storage mode does not support image upgrade in this version")
```
chris-sun-star, line 199: is there any restrictions for observer upgrade in shared storage, which version will support, I think observer in shared storage is more like a stateless service

api/v1alpha1/oblogservicecluster_webhook.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, errors.New("logService is required")
+	}
+	if r.Spec.LogService.Image == "" {
+		return nil, errors.New("logService.image is required")
+	}
+	if r.Spec.ObjectStoreURL.BucketURL == "" {
+		return nil, errors.New("objectStoreUrl.bucketURL is required")
+	}
+	if r.Spec.ObjectStoreURL.SecretRef.Name == "" {
+		return nil, errors.New("objectStoreUrl.secretRef.name is required")
```
chris-sun-star, line 114: It's better to check the secret's existence as well.

internal/cmds/helper/logservice_start.go
```diff
@@ -0,0 +1,205 @@
...
+	// The daemon has forked; find and monitor the actual daemon process
+	return monitorDaemon()
+}
+
+// monitorDaemon polls for the oblogservice daemon process and exits if it dies.
+// It also forwards SIGTERM/SIGINT to the daemon PID for graceful shutdown.
+func monitorDaemon() error {
+	// Wait for daemon to appear (up to 30s for slow startup)
+	var daemonPid int
+	for range 30 {
```
chris-sun-star: is 30 seconds enough for logservice to startup, I'm not very sure about the start process, but for observer we've encountered some problems on slow disk, so the timeout value is pretty large.

internal/cmds/helper/start.go
```diff
@@ -147,14 +166,20 @@ func startOBServerWithParam() error {
...
+	modeFlag := ""
+	if deployMode == "shared_storage" {
+		modeFlag = "-m shared_storage "
+	}
+
 	var cmd string
 	svcIP := os.Getenv("SVC_IP")
 	if standalone != "" && obv.Cmp(MinStandaloneVersion) >= 0 {
-		cmd = fmt.Sprintf("cd %s && %s/bin/observer --nodaemon --appname %s --cluster_id %s --zone %s --devname lo -p %d -P %d -d %s/store -l info -o config_additional_dir=%s/store/etc,%s", DefaultHomePath, DefaultHomePath, clusterName, clusterId, zoneName, DefaultSqlPort, DefaultRpcPort, DefaultHomePath, DefaultHomePath, optStr)
+		cmd = fmt.Sprintf("cd %s && %s/bin/observer %s--nodaemon --appname %s --cluster_id %s --zone %s --devname lo -p %d -P %d -d %s -l info -o config_additional_dir=%s/store/etc,%s", DefaultHomePath, DefaultHomePath, modeFlag, clusterName, clusterId, zoneName, DefaultSqlPort, DefaultRpcPort, dataDir, DefaultHomePath, optStr)
```
chris-sun-star: It's better to leave modeFlag "-m shared_storage" and add the space in the command

internal/resource/obcluster/obcluster_flow.go
```diff
@@ -46,20 +47,24 @@ func genMigrateOBClusterFromExistingFlow(_ *OBClusterManager) *tasktypes.TaskFlo
...
+		tCheckEnvironment,
+		tCheckClusterMode,
+		tCheckAndCreateNs,
+		tCheckAndCreateUserSecrets,
+		tCreateOBZone,
+		tWaitOBZoneBootstrapReady,
+		tBootstrap,
+	}
+	if m.OBCluster.Spec.DeploymentMode == oceanbaseconst.DeploymentModeSharedStorage {
+		tasks = append([]tasktypes.TaskName{tWaitLogServiceReady}, tasks...)
```
chris-sun-star, line 62: is it possible to check logservice in webhook

internal/resource/obcluster/obcluster_manager.go
```diff
@@ -215,11 +215,15 @@ func (m *OBClusterManager) UpdateStatus() error {
 			}
 		}
 
-		// check for upgrade
+		// check for upgrade (shared_storage mode does not support upgrade)
 		if m.OBCluster.Status.Status == clusterstatus.Running {
 			if m.OBCluster.Spec.OBServerTemplate.Image != m.OBCluster.Status.Image {
-				m.Logger.Info("Check obcluster image not match, need upgrade")
-				m.OBCluster.Status.Status = clusterstatus.Upgrade
+				if m.OBCluster.Spec.DeploymentMode == oceanbaseconst.DeploymentModeSharedStorage {
```
chris-sun-star, line 223: I think this is already checked in webhook

internal/resource/observer/observer_task.go
```diff
@@ -390,14 +404,23 @@ func WaitOBServerDeletedInCluster(m *OBServerManager) tasktypes.TaskError {
...
 	for i := 0; i < obcfg.GetConfig().Time.ServerDeleteTimeoutSeconds; i++ {
 		operationManager, err := m.getOceanbaseOperationManager()
 		if err != nil {
 			return errors.Wrapf(err, "Get oceanbase operation manager failed")
 		}
-		observer, err := operationManager.GetServer(m.Ctx, observerInfo)
-		if observer == nil && err == nil {
+		serverExists := false
+		if isSharedStorage {
+			serverExists, err = operationManager.CheckServerExistInDBAOBServers(m.Ctx, observerInfo)
```
chris-sun-star, line 422: Why using a seperate checking logic for sharedStorage

api/v1alpha1/obcluster_webhook.go
Junkrat77: Although shared-storage observers do not own the durable data locally, changing the image can still be a database version upgrade and requires compatibility checks, pre/post upgrade handling, and validation with LogService. These paths have not been validated in this PR, so the restriction is intentional. We do not have a confirmed target release for upgrade support yet, and will track it separately.

api/v1alpha1/oblogservicecluster_webhook.go
Junkrat77: Done. The create webhook now reads the referenced Secret directly from the API server in the same namespace. It rejects a missing Secret and also requires non-empty access_id and access_key fields. Envtest coverage has been added for missing, invalid, and valid Secrets.

internal/cmds/helper/logservice_start.go
Junkrat77: Agreed, 30 seconds may be too short on slow disks. The startup timeout is now configurable through the operator configuration, with a default of 300 seconds. The same deadline covers both launcher exit and daemon discovery. Non-zero launcher exits and timed-out process cleanup are also handled, with unit tests covering these cases.

internal/cmds/helper/start.go
Junkrat77: Done.

internal/resource/obcluster/obcluster_flow.go
Junkrat77: Done. ModeFlag now remains exactly -m shared_storage, and the separator space is added explicitly in each observer command format string.

internal/resource/obcluster/obcluster_manager.go
Junkrat77: Yes, the normal admission path already rejects this change in the webhook. I kept the controller-side check as defense in depth for resources persisted before this validation was introduced, or when admission is bypassed or temporarily unavailable. I added a comment to make that intention explicit.

internal/resource/observer/observer_task.go
Junkrat77: In shared-storage mode, a deleted server remains in __all_server with the cleanup status until _shared_gc_file_expiration_time expires, which is 7 days by default. DBA_OB_SERVERS filters out servers in the cleanup state, so it reflects the logical deletion promptly and prevents the finalizer from being blocked for the retention period. Normal mode continues to use __all_server. I added a comment explaining this distinction, and the shared-storage deletion flow was verified in an actual deployment.

chris-sun-star: OK, understand.

chris-sun-star (approved):

